PR #13, opened 2024-08-02
feat: Make phone numbers optional

### LINKED ISSUE #16: Phone numbers should be optional
Currently, phone numbers are mandatory, however it appears that this is was not the intended design.

tlater-famedly:
Fixes #16

### DIFF
--- a/src/zitadel.rs
+++ b/src/zitadel.rs
@@ -180,15 +180,23 @@ impl Zitadel {
 				.await?;
 		};
 
-		if old.phone != new.phone {
-			self.client
-				.update_human_user_phone(
-					&self.config.famedly.organization_id,
-					user_id.clone(),
-					new.phone.clone(),
-					!self.config.require_phone_verification(),
-				)
-				.await?;
+		match (&old.phone, &new.phone) {
+			(Some(_), None) => {
+				self.client
+					.remove_human_user_phone(&self.config.famedly.organization_id, user_id.clone())
+					.await?;
+			}
+			(_, Some(new_phone)) => {
+				self.client
+					.update_human_user_phone(
+						&self.config.famedly.organization_id,
+						user_id.clone(),
+						new_phone.clone(),
+						!self.config.require_phone_verification(),
+					)
+					.await?;
+			}
+			(None, None) => {}
 		};
 
 		if old.email != new.email {
@@ -317,7 +325,7 @@ struct User {
 	/// The user's LDAP ID
 	ldap_id: String,
 	/// The user's phone number
-	phone: String,
+	phone: Option<String>,
 	/// Whether the user is enabled
 	enabled: bool,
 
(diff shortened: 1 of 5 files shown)


src/zitadel.rs
```diff
@@ -373,7 +373,11 @@ impl User {
 		let preferred_username = read_entry(&entry, &config.ldap.attributes.preferred_username)?;
 		let email = read_entry(&entry, &config.ldap.attributes.email)?;
 		let user_id = read_entry(&entry, &config.ldap.attributes.user_id)?;
-		let phone = read_entry(&entry, &config.ldap.attributes.phone)?;
+		let phone = if let Some(phone_attribute) = config.ldap.attributes.phone {
+			Some(read_entry(&entry, &phone_attribute)?);
+		} else {
+			None;
+		};
```
sirewix, line 380: mb smth like 
```rust
config.ldap.attributes.phone.map(|phone_attribute| read_entry(&entry, &phone_attribute)).transpose()?
```
?

sirewix: Another suggestion just for readability: bind `let attributes = config.ldap.attributes;` shorter `attributes` in `read_entry` calls

tlater-famedly: Yeah, this was done to get it over the line quickly while we were actively testing, didn't want to think through clever option handling.

The attribute binding is a nice idea too. I wonder if something more generic would be possible in general, but I guess we'd need to be able to iterate over struct field names.

sirewix: Yeah, we could make something like 
```rust 
enum Attributes {
  PhoneNumber,
  FirstName,
  ...
}

type LdapAttributesMapping = HashMap<Attributes, String>
```
Then we could iterate over it, but then all attributes would become optional on type level, so I think it's fine as is

src/zitadel.rs
```diff
@@ -409,10 +417,15 @@ impl From<User> for ImportHumanUserRequest {
...
-			}),
+			phone: user.phone.as_ref().map_or_else(
+				|| None,
+				|phone| {
+					Some(Phone {
+						phone: phone.to_owned(),
+						is_phone_verified: !user.needs_phone_verification,
+					})
+				},
+			),
```
sirewix, line 428: Isn't it the same as
```rust
			phone: user.phone.as_ref().map(|phone| Phone {
				phone: phone.to_owned(),
				is_phone_verified: !user.needs_phone_verification,
			}),
```
?

tests/e2e.rs
```diff
@@ -302,6 +310,35 @@ async fn test_e2e_ldaps_starttls() {
...
+
+	let zitadel = open_zitadel_connection().await;
+	let user = zitadel
+		.get_user_by_login_name("no_phone@famedly.de")
+		.await
+		.expect("could not query Zitadel users");
+
+	assert!(user.is_some());
+
+	if let Some(user) = user {
```
sirewix, line 330: Since `user` is asserted to be `is_some`, we can just unwrap here. Or `expect`: `let user = user.expect("no user")`

src/zitadel.rs
tlater-famedly: Yeah, of course it is x) Why did I do that again?

sirewix (approved):

tlater-famedly:
The coverage action cannot pass, since it tries to pull a dependency via ssh, which is no longer possible since this repo is now public.

So merging this without passing tests for now.

We need to rethink the coverage flow as a whole anyway.

PR #17, opened 2024-08-06
Correctly handle binary attributes

### LINKED ISSUE #3: Allow binary/base64 attribute type inputs
LDAP supports encoding attributes as base64 according to this spec: https://www.ibm.com/docs/en/i/7.4?topic=ldif-rfc-2849-input

Such values are marked by having two `:` instead of just the one in LDAP server output. We must assert that we support this, as clients use this.

We should implement tests to ensure this is handled appropriately.

tlater-famedly:
Fixes #3

Binary attributes are synced encoded in base64 to zitadel - zitadel *should* handle that correctly, see https://github.com/zitadel/zitadel/pull/8210.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,5 +1,8 @@
 //! All sync client configuration structs and logic
-use std::path::{Path, PathBuf};
+use std::{
+	fmt::Display,
+	path::{Path, PathBuf},
+};
 
 use anyhow::{bail, Result};
 use ldap_poller::{config::TLSConfig, AttributeConfig, CacheMethod, ConnectionConfig, Searches};
@@ -143,16 +146,16 @@ impl From<LdapConfig> for ldap_poller::Config {
 				page_size: None,
 			},
 			attributes: AttributeConfig {
-				pid: attributes.user_id,
-				updated: attributes.last_modified,
+				pid: attributes.user_id.get_name(),
+				updated: attributes.last_modified.map(AttributeMapping::get_name),
 				additional: vec![],
 				attrs_to_track: vec![
-					attributes.status,
-					attributes.first_name,
-					attributes.last_name,
-					attributes.preferred_username,
-					attributes.email,
-					attributes.phone,
+					attributes.status.get_name(),
+					attributes.first_name.get_name(),
+					attributes.last_name.get_name(),
+					attributes.preferred_username.get_name(),
+					attributes.email.get_name(),
+					attributes.phone.get_name(),
 				],
 			},
 			cache_method: CacheMethod::ModificationTime,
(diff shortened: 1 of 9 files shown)


src/config.rs
```diff
@@ -166,25 +169,61 @@ impl From<LdapConfig> for ldap_poller::Config {
...
 	/// Attribute for the user's unique ID
-	pub user_id: String,
+	pub user_id: AttributeMapping,
 	/// This attribute shows the account status (LDAP = Enabled, accountStatus)
-	pub status: String,
+	pub status: AttributeMapping,
 	/// Marks an account as enabled (LDAP = TRUE, active)
 	pub enable_value: String,
 	/// Marks an account as enabled (LDAP = FALSE, inactive)
 	pub disable_value: String,
```
sirewix, line 188: `StringOrBytes` for these maybe? Since `status` can be binary

src/zitadel.rs
```diff
@@ -430,3 +465,9 @@ impl From<User> for ImportHumanUserRequest {
 		}
 	}
 }
+
+impl Display for User {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		write!(f, "email={}", &self.email)
```
sirewix, line 471: this user string representation is confusing, I'd suggest to `format!` where we need it in-place, if we need it in several places, we can make a function with descriptive name like `user_as_ldapish_string`

tlater-famedly: I mostly wanted a canonical way of displaying the user in log messages; Before this I used `uid`, which turned out to be binary in AD, and flipping every invocation of `user.ldap_id` to `user.email` was annoying, and pointed to needing a `Display` implementation.

As such, I'm not really sure we need to complicate this. Isn't `Display` intended exactly for this, printing user-facing values?

src/config.rs
tlater-famedly: The value of the status attribute can be a binary when received from ldap, but these values can never be binaries since they come from our config file. So I think this represents the actual type of these values better.

src/zitadel.rs
sirewix: And then we also have a bunch of other fields in `User` that we just simply ignore here, and it's a bit meh. I think if we want to display ids we should do `"user {user.id}"`, it's also a bit more explicit in the code, making it clear that only id (or email) is displayed, not the whole user. 

As for "what should we use as id, `ldap_id` or `email`" we can have some method in `User` like `identifier` or `display_identifier` or `printable_identifier` or whatever.

src/config.rs
tlater-famedly: Ok, I've explicitly implemented `PartialEq` so it will compare the bytes as well, in case the status values are in fact binary.

sirewix: what if `enable_value` is some bytestring non-representable as utf? Is it possible to specify invalid utf in yaml/json and make `serde_json` swallow it?

tlater-famedly: I don't really think so. We *could* add explicit support for this, by parsing numbers to `Vec<u8>` if that specific attribute is supposed to be binary, so that users can still represent such values.

That said, in practice this isn't the case for AD or openldap at the moment. I'd say we should hold off on supporting this until we at least have an actual use case.

src/zitadel.rs
tlater-famedly: > And then we also have a bunch of other fields in User that we just simply ignore here

Yeah, if we want a representation that shows all the object's details that's what the `Debug` impl is for.

I'd really rather not have each log potentially log a different attribute as the core identifier, or hardcode it to be represented by a specific attribute in case either that or the exact desired way of representing it changes (we may want to switch to `dn` if group sync ends up being implemented, for example, which is a composite of attributes).

If this is a blocker, we can implement a separate `fmt_for_log`, though I maintain that `Display` at least appears to be the intended trait for that.

sirewix: > Yeah, if we want a representation that shows all the object's details that's what the `Debug` impl is for.

but it's also for debugging, and generally "I don't care how exactly it looks, derive me something I can dump to stdout during debugging".

Can you find an example of a type with `Display` that omits some of the data (most of the data here). I can't think of a single example of a type that would implement `Display` partially, especially for just one field. If I see `user.to_string()` I expect to get user representation in text, `Debug`ish or `Display`ish. Getting just identifier seems unobvious and confusing

> If this is a blocker, we can implement a separate fmt_for_log

I'd suggest to go with this approach, however I'd call it something like `identifier` (`dn` is also identifier)

P.S. The situation in haskell is even [worse](https://harry.garrood.me/blog/down-with-show-part-2/), there's only one `Show` typeclass/trait, which is used for debuging and "displaying", and there's no clear definition of what "displaying" is, as you can have many options: multiline/singleline, compact/pretty/formatted, full/redacted, colored/colorless, syntaxA/syntaxB etc

tlater-famedly: I ended up going with `.log_name()`, that seemed the most simple term for this.

skomski (dismissed):

skomski:
simple_checks should run through. Rustfmt is failing else LGTM

skomski (approved):

PR #18, opened 2024-08-07
feat: Make tls config optional

tlater-famedly:


src/config.rs
```diff
@@ -119,24 +119,32 @@ pub struct LdapConfig {
...
+	pub tls: Option<LdapTlsConfig>,
 }
 
 impl From<LdapConfig> for ldap_poller::Config {
 	fn from(cfg: LdapConfig) -> ldap_poller::Config {
+		let starttls = cfg.tls.clone().is_some_and(|tls| tls.danger_use_start_tls);
+		let no_tls_verify = cfg.tls.clone().is_some_and(|tls| tls.danger_disable_tls_verify);
+		let root_certificates_path = cfg.tls.clone().and_then(|tls| tls.server_certificate);
+		let client_key_path = cfg.tls.clone().and_then(|tls| tls.client_key);
+		let client_certificate_path = cfg.tls.clone().and_then(|tls| tls.client_certificate);
```
sirewix, line 131: Cloning `cfg` 5 times is not efficient, mb something like `cfg.tls.as_ref().and_then(|tls| tls.foo.clone())`? For bool fields there wouldn't be any need to clone as they are `Copy`

sirewix (approved):

src/config.rs
tlater-famedly: Ah, hrm, yeah, makes sense, thanks!

sirewix (approved):

PR #20, opened 2024-08-08
feat!: Implement bitflag for status attribute with multiple disable values

### LINKED ISSUE #12: Improve enable_value and disable_value status
### Description
Right now, we only allow one value for enable or disable status. However in real life, it might happen that we need to figure out if we should create or delete an account based on multiple parameters.

Here is a link of how the userAccountcontrol might look like: https://www.der-windows-papst.de/2016/12/18/active-directory-useraccountcontrol-values/

### Solution
We allow to add multiple values. This means a customer can configure:

```
status: userAccountControl
enable_value: 512, 544, 66048, 66080
disable_value: 2, 514, 546, 66050, 66082
```

In general it will give us more flexibility.

skomski:
Closes #12 

Removes the enable_value and assumes an account is enabled by default.
Expects the status attribute to be parsable as i32 like userAccountControl on AD.
It renames the disable_values to disable_bitflags which are tested against the status attribute like for example ACCOUNTDISABLE (2) for AD.

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -41,9 +41,12 @@ ldap:
       # explicitly. Most of the time it will work either way, but if
       # this is not set any non-UTF8 values will fail to sync.
       is_binary: true
-    status: "shadowInactive"
-    enable_value: 512
-    disable_value: 514
+    # Status flag for the user. This is read as a big-endian integer. Normally userAccountControl in AD.
+    status: "shadowFlag"
+    # Vector of bitmasks that marks the user as disabled. Tested on status.
+    # (for example ACCOUNTDISABLE=0x2 and LOCKOUT=0x10 in AD)
+    # Decimal (or hex) representation of the specific flag mask
+    disable_bitmasks: [0x2, 0x10]
     # Phone numbers are the only optional attribute, if a user does
     # not have a phone number this will be silently ignored
     phone: "telephoneNumber"
--- a/src/user.rs
+++ b/src/user.rs
@@ -1,7 +1,7 @@
 //! User data helpers
 use std::fmt::Display;
 
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use base64::prelude::{Engine, BASE64_STANDARD};
 use ldap_poller::{ldap3::SearchEntry, SearchEntryExt};
 use zitadel_rust_client::{Email, Gender, Idp, ImportHumanUserRequest, Phone, Profile};
@@ -72,8 +72,15 @@ impl User {
 			bail!("missing `{}` values for `{}`", attribute, entry.dn)
 		}
 
-		let enabled = read_entry(&entry, &config.ldap.attributes.status)?
-			!= StringOrBytes::String(config.ldap.attributes.disable_value.clone());
+		let status_as_int = match read_entry(&entry, &config.ldap.attributes.status)? {
+			StringOrBytes::String(status) => status.parse::<i32>()?,
+			StringOrBytes::Bytes(status) => i32::from_be_bytes(
+				status.try_into().map_err(|_| anyhow!("failed to convert to i32 flag"))?,
+			),
+		};
+		let enabled =
+			!config.ldap.attributes.disable_bitmasks.iter().any(|flag| status_as_int & flag != 0);
+
 		let first_name = read_entry(&entry, &config.ldap.attributes.first_name)?;
 		let last_name = read_entry(&entry, &config.ldap.attributes.last_name)?;
 		let preferred_username = read_entry(&entry, &config.ldap.attributes.preferred_username)?;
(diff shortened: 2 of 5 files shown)


config.sample.yaml
```diff
@@ -35,9 +35,10 @@ ldap:
       # explicitly. Most of the time it will work either way, but if
       # this is not set any non-UTF8 values will fail to sync.
       is_binary: true
-    status: "shadowInactive"
-    enable_value: 512
-    disable_value: 514
+    # Status flag for the user. This is read as a big-endian integer. Normally userAccountControl in AD.
+    status: "shadowFlag"
+    # Vector of biflag values that marks the user as disabled. Tested on status. (ACCOUNTDISABLE=2 in AD)
+    disable_bitflags: [2]
```
tlater-famedly, line 41: If we want to match against a bitmask, shouldn't we be parsing a single hex value here instead of a vec?

src/user.rs
```diff
@@ -72,8 +72,15 @@ impl User {
...
-		let enabled = read_entry(&entry, &config.ldap.attributes.status)?
-			!= StringOrBytes::String(config.ldap.attributes.disable_value.clone());
+		let status_as_int = match read_entry(&entry, &config.ldap.attributes.status)? {
+			StringOrBytes::String(status) => status.parse::<i32>()?,
+			StringOrBytes::Bytes(status) => i32::from_be_bytes(
+				status.try_into().map_err(|_| anyhow!("failed to convert to i32 flag"))?,
+			),
+		};
+		let enabled =
+			!config.ldap.attributes.disable_bitflags.iter().any(|flag| status_as_int & flag > 0);
```
tlater-famedly, line 82: I like this, though I suppose we may need to have more configuration options if we ever run into an LDAP schema which doesn't use integer values for this.

config.sample.yaml
skomski: I thought it is easier to configure the invidual flags rather than putting in one bitmask

sirewix: Is this supposed to be the vector of bit indexes? So `[2]` means `0b01` (or `0b001`?)? 
From the code where `&` happens it looks like all the values in this vec are `&`-ed, then I'm not really getting why this is vector

sirewix (dismissed):

config.sample.yaml
skomski: Hope now it is more clear with a second flag as an example LOCKOUT=16. Each flag is tested individual against the status.

sirewix: so, 16 would be 0b10000 mask? or 0b10000000000000000? It seems the first case, but maybe it could be commented with some examples?

sirewix (approved):

config.sample.yaml
sirewix: or maybe rename to `disable_bitmasks` because these seems to be masks, and "bitflags" make me think that the value in the list might be not mask, but a flag, and flag could be represented as the position of the bit (1 << 0 = 0b1, 1 << 1 = 0b10, 1 << 2 = 0b100), and that position could start from the 1 (1 << (1-1) = 0b1, 1 << (2-1) = 0b10, 1 << (3-1) = 0b100), so very ambiguous

skomski: I think it would add more confusion if I would name it bitmask. It should be clear after the given examples and a look of the other flags what is expected https://learn.microsoft.com/en-us/troubleshoot/windows-server/active-directory/useraccountcontrol-manipulate-account-properties.

sirewix: Each value in the list **can** be a bitmask, bitmask **can** include **several** flags. If `disable_bitflags` would not be a list, then there wouldn't be any confusion, as "bitflag**s**" ~ "bitmask". So when I see `s` ending and the type of the value is a list, I strip the `s` and the root of the word should be the type of the elements of the list. Following this logic, the list of `disable_bitflags` should be `disable_bitflag`, stripping the adjective -- just `bitflag`. However bitflag is just 0 or 1, basically boolean, but what is inside `disable_bitflags` list is not that, it is not 0 or 1, it is a bitmask, i.e. many bitflags.

tlater-famedly: I think I'd prefer if this was just a simple bitmask expressed as a hexadecimal integer. Interacting with account status requires understanding bitmasks anyway, having a different format than what people use for other queries is more confusing IMO.

sirewix: > having a different format than what people use for other queries is more confusing

I agree on this, but I don't see where the format is different. I'm merely suggesting to rename a variable to reflect more the underlying values. 

Anyway I don't want this small issue to stall the development process, but if you have time, please gather other opinions from @famedly/workflows

sirewix (dismissed):

sirewix (approved):

config.sample.yaml
skomski: Settled on disable_bitmasks

Issue #4, opened 2024-07-30
Make SSO support separate from the sync tool


tlater-famedly:
Currently, we only sync an identity provider if the SSO feature flag is enabled. This is problematic if a client wants to switch to SSO in the future - flipping on the flag and re-syncing wouldn't work, since this is information only known on the Zitadel side.

Instead, we should always provide an IDP ID, and simply always set up users with support for potential SSO - just disabling the feature in Zitadel.

tlater-famedly:
This ticket is suddenly much more important. For #23 we realized that [passwordless registration enabled us to disable email verification](https://github.com/famedly/ldap-sync/actions/runs/10316262874/job/28558090238?pr=22#step:6:1036), which no longer works if we don't register like that.

Setting the IDP in the request is a valid alternative, so we need to switch to always setting this.

tlater-famedly:
Ok, we have tested this in practice now. It looks like users will not be able to sign in via password if they have an IDP enabled, *even if* external logins are disabled. As such, this is not a good idea, and we shouldn't do it.

PR #22, opened 2024-08-09
fix: Don't set passwordless registration for users

### LINKED ISSUE #23: Remove initialize passwordlss from zitadel user create api call
### Description
This call breaks the "connect account experience". Let's remove it as it is a feature people will not use.

tlater-famedly:
This is *not* enabling users to do their registration without a password, it instead configures setting up registration with authenticators.

Fixes #23 and #4

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,11 +11,15 @@ mod config;
 mod user;
 mod zitadel;
 
-pub use config::{AttributeMapping, Config};
+pub use config::{AttributeMapping, Config, FeatureFlag};
 use zitadel::Zitadel;
 
 /// Run the sync
 pub async fn sync_ldap_users_to_zitadel(config: Config) -> Result<()> {
+	if !config.feature_flags.contains(&FeatureFlag::SsoLogin) {
+		anyhow::bail!("Non-SSO configuration is currently not supported");
+	}
+
 	let cache = read_cache(&config.cache_path).await?;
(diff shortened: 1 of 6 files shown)


tlater-famedly:
Well, that didn't work, looks like we'll need #4 first

tlater-famedly:
Keeping this as a draft until we confirm that this actually works as expected; I fear that this will force users to use SSO.

tlater-famedly:
This now depends on famedly/zitadel-rust-client#9

src/lib.rs
```diff
@@ -11,11 +11,15 @@ mod config;
...
 mod zitadel;
 
-pub use config::{AttributeMapping, Config};
+pub use config::{AttributeMapping, Config, FeatureFlag};
 use zitadel::Zitadel;
 
 /// Run the sync
 pub async fn sync_ldap_users_to_zitadel(config: Config) -> Result<()> {
+	if !config.feature_flags.contains(&FeatureFlag::SsoLogin) {
+		anyhow::bail!("Non-SSO configuration is currently not supported");
```
sirewix, line 20: Is it going to be supported at all? Otherwise having this flag is pointless

tlater-famedly: Yes, but we need to wait for zitadel to add support for it. See the commit message.

sirewix (approved):

PR #36, opened 2024-08-19
Lock the versions of various tooling used in testing

tlater-famedly:
A recent change in the zitadel-tools upstream started requiring go 1.21, which broke our test suite. This should fix it, and I changed all our tooling to use fixed versions while I was at it (except stuff installed on debian, but that should be reasonably stable) - I intended to do that earlier, but forgot.

### DIFF
--- a/tests/environment/test-setup/Dockerfile
+++ b/tests/environment/test-setup/Dockerfile
@@ -1,4 +1,4 @@
-FROM bitnami/openldap:latest
+FROM bitnami/openldap:2.5.18
 
 USER root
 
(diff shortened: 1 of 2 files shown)


tests/environment/test-setup/Dockerfile
```diff
@@ -1,9 +1,12 @@
...
+FROM bitnami/openldap:2.5.18
 
 USER root
 
+RUN echo 'deb http://deb.debian.org/debian bookworm-backports main' >> /etc/apt/sources.list
+
 RUN apt-get update && apt-get upgrade -y && \
-	apt-get install --yes curl golang-go jq && \
+	apt-get install --yes curl jq && \
+	apt-get install --yes -t bookworm-backports golang-go && \
```
sirewix, line 9: Why do we need older go compiler?

sirewix (approved):

tests/environment/test-setup/Dockerfile
tlater-famedly: Newer, backports come from unstable. The zitadel tools package depends on go 1.21 now, but debian bookwork packages 1.20.

PR #37, opened 2024-08-19
feat: Add support for configuration through env var

### LINKED ISSUE #33: Support configuration through environment variables
There's a chance we will need to run this on kubernetes, so we should support environment variables for configuration.

mzaniolo:
Closes #33


src/config.rs
```diff
@@ -9,10 +9,19 @@ use ldap_poller::{config::TLSConfig, AttributeConfig, CacheMethod, ConnectionCon
...
 
+/// App prefix for env var configuration
+const ENV_VAR_CONFIG_PREFIX: &str = "FAMEDLY_LDAP_SYNC";
+
 impl Config {
-	/// Read the config from a file
-	pub async fn from_file(path: &Path) -> Result<Self> {
-		let config: Config = serde_yaml::from_slice(&tokio::fs::read(path).await?)?;
+	/// Create new config from file and env var
+	pub fn new(path: &Path) -> Result<Self> {
```
tlater-famedly, line 17: Let's not rename this - `::new` functions are usually expected to be quite simple (usually just assigning args to struct fields), this invokes IO and can fail.

tlater-famedly (changes requested):
Looks good so far, assuming we finish the detail from https://github.com/famedly/ldap-sync/issues/33#issuecomment-2296777649

src/config.rs
```diff
@@ -9,10 +9,19 @@ use ldap_poller::{config::TLSConfig, AttributeConfig, CacheMethod, ConnectionCon
...
+		let config_builder = config::Config::builder()
+			.add_source(config::File::from(path))
+			.add_source(config::Environment::with_prefix(ENV_VAR_CONFIG_PREFIX).separator("__"));
```
sirewix, line 20: If I understand correctly, if [`config::Environment::list_separator`](https://docs.rs/config/latest/config/struct.Environment.html#method.list_separator) is not set, it's impossible to parse lists? The `config` documentation is a bit lacking, can you add some examples in the readme explaining how to configure via env vars? (especially `Vec` Parameters, including `Set<FeatureFlag>`).

src/config.rs
```diff
@@ -322,6 +334,15 @@ mod tests {
...
 	}
 
+	fn create_config_file(dir: &Path) -> PathBuf {
+		let file_path = dir.join("config.yaml");
+		let mut config_file = File::create(&file_path).expect("failed to create config file");
+		config_file
+			.write_all(EXAMPLE_CONFIG.as_bytes())
+			.expect("Failed to write config file content");
+		file_path
+	}
```
sirewix, line 344: I think with this new approach we can just store `EXAMPLE_CONFIG` as a file somewhere in the repo, to avoid creating it in tempdir

sirewix: Ah, I see you compare it with static `Config`, ok, that makes sense then

src/config.rs
tlater-famedly: We should also add a test case for the one list arg we have (though I still wonder if we shouldn't just use a hexadecimal number for that setting).

mzaniolo: Indeed it was not possible to set lists with the code the way it was. Now it should work. I added tests that cover that part. 
Also, I encountered a small problem, if we try to set a list using empty a string on the env vars it fails. I couldn't solve that so I made added a default for this lists and for setting it as empty we just don't create the variable. There is a note on the readme about that.

README.md
```diff
@@ -33,6 +33,26 @@ Or alternatively, without `docker compose`:
...
+### Configuration
+
+The ldap-sync expects a configuration file located at `./config.yaml`. The default path can be changed setting the new path to the environment variable `FAMEDLY_LDAP_SYNC_CONFIG`.
+
+Also, individual configuration items and the whole configuration can be set using environment variables. For example, the following configuration could be set like this:
+
+Configuration yaml
+```yaml
+ldap:
+          url: ldap://localhost:1389
```
sirewix, line 45: I'm sorry, but 8 space tabs look ugly, we mostly use 2: [.editorconfig:12](https://github.com/famedly/ldap-sync/blob/9dfabce2a41e2749713e545aeee716810574557f/.editorconfig#L12)

sirewix (dismissed):

sirewix (approved):

src/config.rs
tlater-famedly: Hmmm, the value should still be mandatory in the config file. This is a bit awkward, is there really no better solution?

tlater-famedly: I guess an empty list means no values end up marking a user as disabled, which is quite intuitive. Yeah, we can keep that behavior.

sirewix: If that list cannot be empty, then we should have used something like [`nonempty`](https://docs.rs/nonempty/latest/nonempty/) in the first place

tlater-famedly (approved):

sirewix (approved):

PR #40, opened 2024-08-21
feat: Add dry-run flag

### LINKED ISSUE #28: Implement `dry run` feature
It would be good to have a command line option `--dry-run` which would only log the added,updated,deleted users in LDAP but would abort before interacting with zitadel and commiting them to the cache.

tlater-famedly:
Fixes #28


nikzen:
This probably needs some documentation of how to use the dry-run mode

tlater-famedly:
> This probably needs some documentation of how to use the dry-run mode

Fair point, I didn't want to put it in the sample config, but that's where we currently document all our config settings. Would be nice if we could generate some docs from our `config.rs`.

sirewix (approved):

PR #41, opened 2024-08-22
feat: Add deactivate only mode

### LINKED ISSUE #38: Deactivate only mode
### Description
We got a request, if we can implement a "deactivate only" mode. In this mode, it will only check for deactivated users and will not provision new users.

### User story
As a Famedly customer, I want to use ldap-sync for the sole purpose of deactivating users in Famedly that have already been deactivated in my LDAP.

### Steps to do
- Add a feature flag
- Only search for deactivated users per the configured `disable_bitmasks`.
- Find matching users in Zitadel
- Disable matching users

mzaniolo:
Closes #38


sirewix (approved):

src/lib.rs
```diff
@@ -48,9 +50,12 @@ pub async fn sync_ldap_users_to_zitadel(config: Config) -> Result<()> {
 
 	sync_handle.await??;
 
-	zitadel.import_new_users(added).await?;
 	zitadel.update_users(changed).await?;
```
emgrav, line 53: Won't this still mean that a disabled user being enabled will also propagate from LDAP to Zitadel?

src/config.rs
```diff
@@ -279,6 +279,8 @@ pub enum FeatureFlag {
 	VerifyPhone,
 	/// If set, only log changes instead of writing anything
 	DryRun,
+	///If only deactivated user should be sync
```
emgrav, line 282: Just a tiny nit
```suggestion
	/// If only deactivated users should be synced
```

emgrav (changes requested):
I have one question about the functionality and a tiny nit about a comment, but looks quite good otherwise.

src/lib.rs
mzaniolo: No, inside zitadel.rs we have a differentiation between the changes to enable and disable (line 91). I have another check for this feature flag on the zitadel.rs to handle this case. However, I just extended my test case to also cover this scenario

emgrav (approved):
LGTM

src/lib.rs
```diff
@@ -24,6 +24,8 @@ pub async fn sync_ldap_users_to_zitadel(config: Config) -> Result<()> {
 	let zitadel = Zitadel::new(&config).await?;
 	let (mut ldap_client, ldap_receiver) = Ldap::new(config.clone().ldap.into(), cache);
 
+	let deactivate_only = config.feature_flags.contains(&FeatureFlag::DeactivateOnly);
```
tlater-famedly, line 27: Small nit: We could have a `config.disable_only()` so we don't need to do this `.contains` shenanigans all the time.

sirewix: We do have such function for some other feature flag. I'd say we should add some generic method to config:
```rust
pub fn is_feature_enabled(feature_flag: FeatureFlag) -> bool {...}
```
However the ".contains shenanigans" are quite readable, so whatever, but let's make it consistent with other "is feature X enabled" expressions, explicit or hidden behind some function

sirewix:
Can you add some lines to the readme and/or config sample on how cache behaves when this mode is enabled. From what I see the cache is written as if all changes were processed, so it wouldn't be possible to run with `disable_only` and then without and get every change processed that wasn't processed during `disable_only` run.

mzaniolo:
> Can you add some lines to the readme and/or config sample on how cache behaves when this mode is enabled. From what I see the cache is written as if all changes were processed, so it wouldn't be possible to run with `disable_only` and then without and get every change processed that wasn't processed during `disable_only` run.

Well, I think we should first be ask how we want the cache to behave when deactivate only is set. The issue doesn't say anything about it. Does anyone know how to answer that? Maybe @nikzen ?

sirewix (approved):

Issue #55, opened 2024-09-06
Users with non-canonical phone numbers cannot be synced


tlater-famedly:
Currently, users with phone numbers that Zitadel does not accept will fail to sync, as Zitadel validates phone numbers before importing users.

In practice, it seems most users do not specify valid phone numbers in LDAP, so we need to do something about this besides telling people to specify valid phone numbers.

The most reasonable approach is probably to check error messages to see if a user failed because of an invalid phone number, and if that happens immediately try to re-import the user, but without a phone number; Writing our own regex or copying Zitadels' risks drifting away from their exact implementation, which may cause failed syncs in the future, so even if we implement some kind of local check to prevent sending as many requests we should still do some kind of double sync.

emgrav:
We could also add a configurable country code and attempt to canonicalize the phone numbers with the configured country code. While there will be edge cases, these should be limited to cases where 1) A user has a phone number from another country than their organization is in, and 2) This international phone number is missing a country code. While there will probably be quite a bit of the first, both of these being true at once seem unlikely.

tlater-famedly:
Yeah, that is probably sensible if we want to ensure that all users that have phone numbers set actually get them synced; still, even if we canonicalize the phone numbers, we should handle cases where Zitadel doesn't like the phone numbers. They are explicitly optional, users should not fail to sync because their phone numbers are in an unexpected format.

tlater-famedly:
We discussed this in the sync - @nikzen suggests that the phone number may even be internal to e.g. a hospital, rather than a "real" phone number, so we should not try to add a prefix in general. Simply adding a retry for this should be sufficient.

PR #62, opened 2024-09-16
Retry sync for users with invalid phone numbers

jannden:
Closes: #55

### DIFF
--- a/src/zitadel.rs
+++ b/src/zitadel.rs
@@ -47,16 +47,46 @@ impl Zitadel {
 			zitadel_client,
 		})
 	}
+
 	/// Import a list of new users into Zitadel
 	pub(crate) async fn import_new_users(&self, users: Vec<User>) -> Result<()> {
 		for user in users {
 			let zitadel_user =
 				user.to_zitadel_user(&self.feature_flags, &self.zitadel_config.idp_id);
-			let sync_status = self.import_user(&zitadel_user).await;
+			let status = self.import_user(&zitadel_user).await;
 
-			if let Err(error) = sync_status {
-				tracing::error!("Failed to sync user `{}`: {:?}", zitadel_user.log_name(), error);
-			};
+			if let Err(error) = status {
+				tracing::error!(
+					"Failed to sync-import user `{}`: {:?}",
+					zitadel_user.log_name(),
+					error
+				);
+
+				if Self::is_invalid_phone_error(error) {
+					let zitadel_user = ZidatelUser {
+						user_data: User { phone: None, ..zitadel_user.user_data },
+						..zitadel_user
+					};
+
+					let retry_status = self.import_user(&zitadel_user).await;
+
+					match retry_status {
+						Ok(_) => {
+							tracing::info!(
+								"Retry sync-import succeeded for user `{}`",
+								zitadel_user.log_name()
+							);
+						}
+						Err(retry_error) => {
+							tracing::error!(
+								"Retry sync-import failed for user `{}`: {:?}",
+								zitadel_user.log_name(),
+								retry_error
+							);
+						}
+					}
+				}
+			}
 		}
 
 		Ok(())
@@ -144,7 +174,30 @@ impl Zitadel {
 				let status = self.update_user(&old, &new).await;
 
 				if let Err(error) = status {
-					tracing::error!("Failed to update user `{}`: {:?}", new.log_name(), error);
+					tracing::error!("Failed to sync-update user `{}`: {:?}", new.log_name(), error);
+
+					if Self::is_invalid_phone_error(error) {
+						let new =
+							ZidatelUser { user_data: User { phone: None, ..new.user_data }, ..new };
+
+						let retry_status = self.update_user(&old, &new).await;
+
+						match retry_status {
+							Ok(_) => {
+								tracing::info!(
+									"Retry sync-update succeeded for user `{}`",
+									new.log_name()
+								);
+							}
+							Err(retry_error) => {
+								tracing::error!(
+									"Retry sync-update failed for user `{}`: {:?}",
+									new.log_name(),
+									retry_error
+								);
+							}
+						}
+					}
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/zitadel.rs
```diff
@@ -50,13 +50,34 @@ impl Zitadel {
...
-			let zitadel_user =
+			let mut zitadel_user =
 				user.to_zitadel_user(&self.feature_flags, &self.zitadel_config.idp_id);
 			let sync_status = self.import_user(&zitadel_user).await;
 
 			if let Err(error) = sync_status {
 				tracing::error!("Failed to sync user `{}`: {:?}", zitadel_user.log_name(), error);
-			};
+
+				if error.to_string().contains("invalid ImportHumanUserRequest_Phone") {
```
tlater-famedly, line 60: That's not particularly nice, the error in this case is a `Status` struct which should have more structured info than a string. Have you looked into how much that can be parsed instead?

tests/e2e.rs
```diff
@@ -399,6 +399,83 @@ async fn test_e2e_no_phone() {
...
+#[test(tokio::test)]
+#[test_log(default_log_filter = "debug")]
+async fn test_e2e_sync_invalid_phone() {
+	let mut ldap = Ldap::new().await;
+	ldap.create_user(
+		"John",
+		"Good Phone",
+		"Johnny1",
+		"good_phone@famedly.de",
+		Some("+12015550123"),
```
tlater-famedly, line 411: I think we can simplify this test case by removing the good branch, other tests already assert that we can import phone numbers.

src/zitadel.rs
```diff
@@ -145,6 +166,25 @@ impl Zitadel {
 
 				if let Err(error) = status {
 					tracing::error!("Failed to update user `{}`: {:?}", new.log_name(), error);
+
+					if error.to_string().contains("invalid UpdateHumanPhoneRequest") {
```
tlater-famedly, line 170: We really shouldn't copy-paste code like this.

tlater-famedly (changes requested):

src/zitadel.rs
```diff
@@ -145,6 +166,25 @@ impl Zitadel {
...
+						let mut new = new.clone();
+						new.user_data.phone = None;
```
sirewix, line 172: [`assign`](https://docs.rs/assign/latest/assign/) maybe?

src/zitadel.rs
jannden: It's possible to go one level deeper, so I updated that. For reference, this is the error when importing user:
```
Status {
        code: InvalidArgument,
        message: "invalid ImportHumanUserRequest.Phone: embedded message failed validation | caused by: invalid ImportHumanUserRequest_Phone.Phone: value does not have prefix \"+\"",
        metadata: MetadataMap {/* ... */},
        source: None,
}
```

src/zitadel.rs
jannden: I changed it to the [struct update syntax](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax), let me know if you insist on the `assign` macro instead.

src/zitadel.rs
jannden: I exported the check to a function.

tests/e2e.rs
jannden: The good branch is there so that we test updating the good branch to a bad number. I will change the naming to be clearer.

src/zitadel.rs
```diff
@@ -47,16 +47,46 @@ impl Zitadel {
...
-			};
+			if let Err(error) = status {
+				tracing::error!(
+					"Failed to sync-import user `{}`: {:?}",
+					zitadel_user.log_name(),
+					error
+				);
+
+				if Self::is_invalid_phone_error(error) {
+					let zitadel_user = ZidatelUser {
```
emgrav, line 66: Looks like we missed this in the previous review, this struct is named `ZidatelUser` :D We should fix that, but I don't think it needs to be this PR as I'd say it's ready to merge.

sirewix (approved):

emgrav (approved):

tlater-famedly (approved):

src/zitadel.rs
```diff
@@ -47,16 +47,46 @@ impl Zitadel {
...
-				tracing::error!("Failed to sync user `{}`: {:?}", zitadel_user.log_name(), error);
-			};
+			if let Err(error) = status {
+				tracing::error!(
+					"Failed to sync-import user `{}`: {:?}",
+					zitadel_user.log_name(),
+					error
+				);
+
+				if Self::is_invalid_phone_error(error) {
```
tlater-famedly, line 65: The code is still incredibly similar between the two functions, I think there's a single word difference between the two implementations. Can't this logic be shared too?

Personally I wonder if this logic should be lifted up one abstraction layer, returning an `Error` with an enum value we can process in the parent function call, which then rearranges another set of imports with adjusted phone numbers. Not that I've fully thought through a good architecture for this, but needing this many extremely specific utility functions feels like a code smell.

I suppose it's fine to hold off on thinking through things too much for now since we do need this feature, but we should redesign this in the future.

src/zitadel.rs
tlater-famedly: I don't quite follow, the name seems appropriate?

emgrav: It's "Zidatel" instead of "Zitadel"

jannden: Good catch! I couldn't see it either 😂

src/zitadel.rs
jannden: From what I understand, you are trying not go against DRY.

Let's think about it in terms of pseudocode. Currently:

1. ask zitadel-rust-client to **import** a user
2. call the local function to check if returned error is about phone numbers
3. if so, try to **import** the user without phone number

Then the other place we go over this procedure, instead of **import** it is **update**. So the only step that repeats is stripping the user of the phone number.

Looks clean enough to me - at least at this point. If we do decide to redesign this in the future, I will be more than happy to assist if I can.

Issue #53, opened 2024-09-05
Get rid of "cache" concept


tlater-famedly:
The "cache" is a holdover from [ldap-poller](https://github.com/famedly/ldap-poller), and used to keep track of what changed since the last run of the tool, so that we know when to send update and deletion requests instead of import requests.

While this works for now, and is fine as long as ldap-sync is the only tool performing syncs and the "cache" is persisted, the databases will go out of sync the moment this is not true.

The name "cache" is therefore a big misnomer, since the name implies it can be safely removed at a performance penalty, when in reality doing so breaks the tool. We also currently don't have a good way to get things back into sync when issues (e.g. failed requests) occur because ldap-poller doesn't have an interface to bust the "cache".

We should move away from this.

tlater-famedly:
Waiting for zitadel-rust-client#15

PR #64, opened 2024-09-18
Add CSV Source

### LINKED ISSUE #52: Add CSV source
Blocked by #49 (This PR adds a source trait which we should implement for a CSV source)

### Description
We want to add an additional data source to the sync tool, specifically CSV (or some other form of flatfile source). We have a need for this quite soon, as it would significantly simplify the process of bootstrapping the demo environment for Operations and Marketing.

### User story
As a Famedly customer, I want to be able to provide a csv file of user details to import into Famedly so I can easily provision my list of users in bulk.

### Requirements
* CSV format (needs to also define the format so we can communicate this to Marketing who will use it, as well as document it in the readme)

### Steps to implement
* Add a new implementation of the source trait that reads a CSV file from disk

jannden:
Closes #52 , https://github.com/famedly/product-management/issues/2038

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,8 @@ url = "2.5.2"
 uuid = { version = "1.10.0", features = ["v5"] }
 zitadel-rust-client = { git = "https://github.com/famedly/zitadel-rust-client", version = "0.1.0" }
 wiremock = "0.6.2"
+csv = "1.3.0"
+tempfile = "3.12.0"
 
 [dependencies.tonic]
 version = "*"
@@ -134,4 +136,3 @@ unwrap_used = "warn"
 used_underscore_binding = "warn"
 useless_let_if_seq = "warn"
 verbose_file_reads = "warn"
-
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -107,3 +107,11 @@ sources:
     scope: "openid read-maillist"
     # Grant type
     grant_type: client_credentials
+
+  # Configuration for the CSV source - reads a CSV file
+  # and creates **new** users in Famedly's Zitadel.
+  # Expected structure of the CSV file is as follows:
+  # email,first_name,last_name,phone
+  csv:
+    # Path to the CSV file to read from.
+    file_path:  ./tests/environment/files/test-users.csv
--- a/src/sources/csv.rs
+++ b/src/sources/csv.rs
@@ -1,0 +1,266 @@
+//! CSV source for syncing with Famedly's Zitadel.
+
+use std::{fs, path::PathBuf};
+
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use csv::Reader;
+use serde::Deserialize;
+
+use super::Source;
+use crate::{user::User, zitadel::SourceDiff};
+
+/// CSV Source
+pub struct CsvSource {
+	/// CSV Source configuration
+	csv_config: CsvSourceConfig,
+}
+
+#[async_trait]
+impl Source for CsvSource {
+	fn get_name(&self) -> &'static str {
+		"CSV"
+	}
+
+	async fn get_diff(&self) -> Result<SourceDiff> {
+		let new_users = self.read_csv()?;
+		// TODO: Implement changed and deleted users
+		// Holding off on this until we get rid of the cache concept
+		// https://github.com/famedly/ldap-sync/issues/53
+		return Ok(SourceDiff { new_users, changed_users: vec![], deleted_user_ids: vec![] });
+	}
+}
+
+impl CsvSource {
+	/// Create a new CSV source
+	pub fn new(csv_config: CsvSourceConfig) -> Self {
+		Self { csv_config }
+	}
+
+	/// Get list of users from CSV file
+	fn read_csv(&self) -> Result<Vec<User>> {
+		let file_path = &self.csv_config.file_path;
+		let file = fs::File::open(&self.csv_config.file_path)
+			.context(format!("Failed to open CSV file {}", file_path.to_string_lossy()))?;
+		let mut reader = Reader::from_reader(file);
+		Ok(reader
+			.deserialize()
+			.map(|r| r.inspect_err(|x| tracing::error!("Failed to deserialize: {x}")))
+			.filter_map(Result::ok)
+			.map(CsvData::to_user)
+			.collect())
+	}
+}
+
+/// Configuration to get a list of users from a CSV file
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub struct CsvSourceConfig {
+	/// The path to the CSV file
+	pub file_path: PathBuf,
+}
+
+/// CSV data structure
+#[derive(Debug, Deserialize)]
+struct CsvData {
+	/// The user's email address
+	email: String,
+	/// The user's first name
+	first_name: String,
+	/// The user's last name
+	last_name: String,
+	/// The user's phone number
+	phone: String,
+}
+
+impl CsvData {
+	/// Convert CsvData to User data
+	fn to_user(csv_data: CsvData) -> User {
+		User {
+			email: csv_data.email.clone().into(),
+			first_name: csv_data.first_name.into(),
+			last_name: csv_data.last_name.into(),
+			phone: if csv_data.phone.is_empty() { None } else { Some(csv_data.phone.into()) },
+			preferred_username: csv_data.email.clone().into(),
+			external_user_id: csv_data.email.into(),
+			enabled: true,
+		}
+	}
+}
+
+/// Helper module for unit and e2e tests
+pub mod test_helpers {
+	use std::fs::write;
+
+	use anyhow::Result;
+	use tempfile::NamedTempFile;
+
+	use crate::Config;
+
+	/// Prepare a temporary CSV file with the given content and update the
+	/// config to use it as the CSV source file.
+	pub fn temp_csv_file(config: &mut Config, csv_content: &str) -> Result<NamedTempFile> {
+		let temp_file = NamedTempFile::new()?;
+		write(temp_file.path(), csv_content)?;
+
+		if let Some(csv) = config.sources.csv.as_mut() {
+			csv.file_path = temp_file.path().to_path_buf();
+		}
+
+		Ok(temp_file)
+	}
+}
+
+#[cfg(test)]
+mod tests {
+
+	use indoc::indoc;
+
+	use super::*;
+	use crate::{user::StringOrBytes, Config};
+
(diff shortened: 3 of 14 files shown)


jannden:
The requirements were clarified and confirmed with the only stakeholder at the moment - the Marketing team. Task specified https://github.com/famedly/ldap-sync/issues/52
Namely:
- only importing new users,
- skipping import for users that already exist,
- reading CSV file from disk,
- CSV data format: `email,first_name,last_name,preferred_username,phone,enabled`

src/sources/csv.rs
```diff
@@ -0,0 +1,261 @@
...
+	fn read_csv(&self) -> Result<Vec<User>> {
+		let file_path = &self.csv_config.file_path;
+		let file = fs::File::open(&self.csv_config.file_path)
+			.context(format!("Failed to open CSV file {}", file_path.to_string_lossy()))?;
+		let mut reader = Reader::from_reader(file);
+		let users: Result<Vec<CsvData>, _> = reader.deserialize().collect();
+		let users: Result<Vec<User>, _> = users
+			.context("Failed to deserialize CSV data")
+			.map(|users| users.into_iter().map(CsvData::to_user).collect());
+		users
```
sirewix, line 47: Small suggestion to avoid unnecessary bindings:
```rust
		Ok(reader
			.deserialize()
			.collect::<Result<Vec<_>, _>>()
			.context("Failed to deserialize CSV data")?
			.into_iter()
			.map(CsvData::to_user)
			.collect())
```
Or to actually ignore deserialization errors:
```rust
		Ok(reader
			.deserialize()
			.map(|r| r.inspect_err(|x| tracing::error!("Failed to deserialize: {x}")))
			.filter_map(Result::ok)
			.map(CsvData::to_user)
			.collect())
```
Or if you still want to keep it splitted, we should unwrap IMO instead of `.map` on `Result`:
```rust
		let users: Vec<CsvData> = reader
			.deserialize()
			.collect::<Result<Vec<_>, _>>()
			.context("Failed to deserialize CSV data")?;
		Ok(users.into_iter().map(CsvData::to_user).collect())
```

src/sources/csv.rs
```diff
@@ -0,0 +1,261 @@
...
+			phone: if csv_data.phone.is_empty() { None } else { Some(csv_data.phone.into()) },
+			enabled: csv_data.enabled,
+			preferred_username: csv_data.preferred_username.clone().into(),
+			ldap_id: csv_data.preferred_username.into(),
+		}
+	}
+}
+
+/// Helper module for unit and e2e tests
+pub mod test_helpers {
```
sirewix, line 91: `#[cfg(test)]` maybe?

src/sources/csv.rs
```diff
@@ -0,0 +1,261 @@
...
+
+#[async_trait]
+impl Source for CsvSource {
+	fn get_name(&self) -> &'static str {
+		"CSV"
+	}
+
+	async fn get_diff(&self) -> Result<SourceDiff> {
+		let new_users = self.read_csv()?;
+		return Ok(SourceDiff { new_users, changed_users: vec![], deleted_user_ids: vec![] });
```
sirewix, line 27: Continuing on [the past discussion](https://github.com/famedly/ldap-sync/pull/49#discussion_r1754790510), I think we should at least add a comment that this only implements adding users.

Potentially this could work for changing users as well, as for deleting it would require different csv table format

src/sources/csv.rs
sirewix: We should add `#![cfg(test)]` to those (as in https://github.com/famedly/backend-system-tests/blob/ef787dc3d6a882e18f1333abf9c5ee73e7935151/tests/zitadel.rs#L1)

src/sources/csv.rs
tlater-famedly: Agreed. I think we should also double check with marketing whether this is actually the *desired* behavior. Could well be that they didn't consider the edge case of "some existing user on an instance matches with the ones we want to sync but has outdated data".

Of course, if we want to update users we need to first query Zitadel, which is basically #53, and in fact updating users will make #53 significantly easier to make generic. We can merge this PR either way, but we should add a TODO if we want to change the behavior.

src/sources/csv.rs
```diff
@@ -0,0 +1,261 @@
...
+#[derive(Debug, Deserialize)]
+struct CsvData {
+	/// The user's email address
+	email: String,
+	/// The user's first name
+	first_name: String,
+	/// The user's last name
+	last_name: String,
+	/// The user's preferred username
+	preferred_username: String,
```
tlater-famedly, line 68: We basically only record this to add it to metadata to remember it since it's available on ldap, it's never used in practice. Do we actually need to define it for the CSV source?

src/sources/csv.rs
```diff
@@ -0,0 +1,261 @@
...
+	/// The user's phone number
+	phone: String,
+	/// Whether the user is enabled
+	enabled: bool,
```
tlater-famedly, line 72: This should always be true, disabled users are simply deleted from Zitadel

src/user.rs
```diff
@@ -39,7 +39,7 @@ impl User {
 
 /// Crate-internal representation of a Zitadel user
 #[derive(Clone, Debug)]
-pub struct ZidatelUser {
+pub struct ZitadelUser {
```
tlater-famedly, line 42: Let's put this in a separate commit, this just muddies the blame.

tests/environment/files/test-users.csv
```diff
@@ -0,0 +1,5 @@
+email,first_name,last_name,preferred_username,phone,enabled
+john.doe@example.com,John,Doe,jdoe,"+1111111111",true
+jane.smith@example.com,Jane,Smith,jasmith,"+2222222222",true
+alice.johnson@example.com,Alice,Johnson,ajohnson,,true
+bob.williams@example.com,Bob,Williams,bwilliams,"+4444444444",true
```
tlater-famedly, line 5: Nit: Another EOF newline ;) The precommit hook checks for this, maybe we should add a check to CI as well.

src/sources/csv.rs
tlater-famedly: May need to ask product for specifics here.

src/user.rs
emgrav: This is also already corrected on main

tlater-famedly: Yep, spotted that after, ta @emgrav

src/sources/csv.rs
jannden: rustc doesn't like that and says `#![no_std]` above the test_helpers module and says inner attributes are usually found at the beginning of source files

src/sources/csv.rs
jannden: I added an explanatory comment.

src/sources/csv.rs
sirewix: I mean we can put `#![cfg(test)]` to the top of the `test/*.rs` modules, then they should compile if `test_helpers` is feature gated by `#[cfg(test)]`

config.sample.yaml
```diff
@@ -107,3 +107,11 @@ sources:
...
     # Grant type
     grant_type: client_credentials
+
+  # Configuration for the CSV source - reads a CSV file
+  # and creates **new** users in Famedly's Zitadel.
+  # Expected structure of the CSV file is as follows:
+  # email,first_name,last_name,preferred_username,phone,enabled
+  csv:
+    # Path to the CSV file to read from.
+    file_path:  ./tests/environment/files/test-users.csv
```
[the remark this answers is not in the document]
jannden: Done.

Cargo.toml
```diff
@@ -25,6 +25,8 @@ zitadel-rust-client = { git = "https://github.com/famedly/zitadel-rust-client",
 chrono = "0.4.19"
```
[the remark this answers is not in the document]
jannden: Done.

src/sources/csv.rs
jannden: Yes, looks like it's kept for linking LDAP user to Zitadel user and it's quite entangled in the old code, specifically the User struct and everywhere it's used. I will see to what would be the best way to separate it. Probably making it an Option so that LDAP still can use it, but other sources don't have to.

tests/environment/files/test-users.csv
jannden: Done.

src/sources/csv.rs
jannden: The thing is that those modules have to be re-exported from lib.rs, which gives: found an item that was configured out. Am I missing something?

src/sources/csv.rs
jannden: Done.

src/sources/csv.rs
jannden: Done. I made the preferred_username being the email address by default if not supplied by LDAP otherwise.

src/sources/csv.rs
sirewix: you're right, that's impossible

emgrav:
UKF wants a similar feature, apparently. See the linked issue above. I think it would make sense to add fields for organization, project, and role.

Sorry about not catching this issue earlier

nikzen:
> UKF wants a similar feature, apparently. See the linked issue above. I think it would make sense to add fields for organization, project, and role.
> 
> Sorry about not catching this issue earlier

@emgrav THis is not required right now for UKF - so we can keep it as it is

sirewix (approved):

src/sources/csv.rs
tlater-famedly: *Technically* I think it'd be best if we didn't set this field at all when the CSV source is used. I'm ok with either implementation though, given we don't actually know why we're setting this field at the moment.

tlater-famedly (approved):

Cargo.toml
```diff
@@ -26,6 +26,8 @@ url = "2.5.2"
 uuid = { version = "1.10.0", features = ["v5"] }
 zitadel-rust-client = { git = "https://github.com/famedly/zitadel-rust-client", version = "0.1.0" }
 wiremock = "0.6.2"
+csv = "1.3.0"
+tempfile = "3.12.0"
```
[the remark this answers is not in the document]
jannden: Looks like `tempfile` already was in [the dependencies](https://github.com/famedly/ldap-sync/blob/main/Cargo.lock#L1505). Weird historically, but all looks good at this point.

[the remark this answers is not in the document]
tlater-famedly: Fair enough, I assume it's a dependency of something.

PR #65, opened 2024-09-18
Release v0.4.0

tlater-famedly:
Had to doctor the changelog a bit; we should be more careful with commit titles.


tlater-famedly:
Superseded by #65

PR #66, opened 2024-09-18
Small fixes and release v0.4.0

emgrav:


sirewix (approved):

sirewix (dismissed):

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+- Add UKT Source support
+- Add CSV Source support
+
+### Bug Fixes
+
+- Enable the native-tls feature for tonic/reqwest
+
+### Refactor
+
+- [**breaking**] Adjust config struct for multi-source trait
```
tlater-famedly, line 18: Many of these aren't really refactors, we should have been a bit more careful reviewing commit titles.

Shall we move them to features to prevent confusing infra?

emgrav: I think it's too late to change now that that PR is merged. iirc git-cliff will just keep regenerating that commit message in newer versions if it doesn't find a match, which it won't if we manually change it

tlater-famedly: I thought it only goes up to the last tag if configured with our git-cliff config, but your call.

sirewix: I often adjust/edit changelog, I don't think it regenerates old ones

PR #68, opened 2024-09-18
ci: Add release workflow with SBOM and attestation

emgrav:


.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+          gitlab_user: ${{ secrets.GITLAB_USER }}
+          gitlab_pass: ${{ secrets.GITLAB_PASS }}
+
+      - name: Caching
+        uses: Swatinem/rust-cache@b8a6852b4f997182bdea832df3f9e153038b5191
+        with:
+          cache-on-failure: true
+          cache-all-crates: true
+
+      - name: Install additional cargo tooling
```
tlater-famedly, line 47: We really need to get this stuff into the rust image itself. Not a blocker for this PR, and I guess I'll be doing that with the CI rework anyway.

.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: release-*
+          path: artifacts
+          merge-multiple: true
+
+      - name: Create release
+        uses: softprops/action-gh-release@79721680dfc87fb0f44dfe65df68961056d55c38
+        with:
+          files: artifacts/*
```
emgrav, line 129: Could be nice to inspect the triggering tag and mark the release as a pre-release depending on whether the tag includes `rc.X`

.github/workflows/release.yml
```diff
@@ -0,0 +1,130 @@
...
+      contents: read
+      attestations: write
+    strategy:
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-latest
+            asset_name: ${{ github.event.repository.name }}-linux-amd64-latest
+          - platform: linux/arm64
+            runner: arm-ubuntu-latest-8core
```
tlater-famedly, line 26: Is this one of the expensive runners or are there no arm runners with unspecified numbers of cores?

tlater-famedly (dismissed):
LGTM, I guess the idea is to use ldap-sync as a test bed for this too, then move it to the shared workflows?

emgrav:
Yep

.github/workflows/release.yml
emgrav: done

sirewix (approved):

mzaniolo (approved):

Issue #35, opened 2024-08-19
TLS tests fail on Mac OS


tlater-famedly:
Apparently Mac OS has very strict requirements for SSL certificates, which we do not meet with the certificates in our test suite: https://support.apple.com/en-us/103769

emgrav:
Which requirements are not met?

tlater-famedly:
I *believe*:

- TLS server certificates must present the DNS name of the server in the Subject Alternative Name extension of the certificate. DNS names in the CommonName of a certificate are no longer trusted.
- TLS server certificates must contain an ExtendedKeyUsage (EKU) extension containing the id-kp-serverAuth OID.
- TLS server certificates must have a validity period of 825 days or fewer (as expressed in the NotBefore and NotAfter fields of the certificate).

We can consider issuing the certificate in the past to fix that last bit, so we don't have to constantly regenerate certificates. *Or* call the certificate generation script during test setup.

tlater-famedly:
It requires having openssl (not just as a library) on the host running the tests, or adding another container to the test runs. So either require more host tooling - which I've avoided so far to simplify getting started as much as possible - or slow down tests a fair bit.

sirewix:
I see, then maybe issuing a cetificate in the past is not such a bad idea. However I'm also fine with adding more tooling on the host or into the container (probably the latter is better). Regarding *only-when-expired* check, that could probably be done with some lock file where we store the issue date (to avoid calling openssl on each run to check the issue date).

tlater-famedly:
> that could probably be done with some lock file where we store the issue date (to avoid calling openssl on each run to check the issue date)

Maybe, if we have a requirement on openssl anyway that shouldn't really be a meaningful overhead, though, *unless* we prevent starting a container altogether, which I don't think is trivial?

sirewix:
> which I don't think is trivial?

I thought you meant having separate container for openssl:

> or adding another container to the test runs

In that case it would be a matter of running `docker run openssl -- keygen blabla` or not running it.

If we add openssl to the container with the test suite, then yeah, no difference ofcourse

emgrav:
We will generate issues before July 1st, 2019 with the server name in the server name extension and not just CommonName.

jannden:
I suspect that there might be something else in play in addition to the aforementioned.

[...]

PR #76, opened 2024-10-24
test: Fix certificate being not trusted on MacOS

mzaniolo:
closes: #35


emgrav (dismissed):
Very nice to have this done, thanks!

.config/starte2e.sh
```diff
@@ -10,6 +10,9 @@ export PATH="${PATH}:/usr/bin"
 touch tests/environment/zitadel/service-user.json
 chmod a+rw tests/environment/zitadel/service-user.json
 
+# Generate certificates
+tests/environment/certs/generate-certs.sh
+
 # Shut down any still running test-setup first
-docker compose --project-directory ./tests/environment down -v test-setup || true
+docker compose --project-directory ./tests/environment down -v || true
```
tlater-famedly, line 17: This is deliberate to avoid having to restart the more expensive services. It saves ~20 seconds per test run on my end. The only thing that sometimes gets stuck between reboots IME is the docker network, we should delete that one explicitly.

.config/starte2e.sh
```diff
@@ -10,6 +10,9 @@ export PATH="${PATH}:/usr/bin"
 touch tests/environment/zitadel/service-user.json
 chmod a+rw tests/environment/zitadel/service-user.json
 
+# Generate certificates
+tests/environment/certs/generate-certs.sh
```
tlater-famedly, line 14: We should consider doing this in the docker container to avoid the openssl dependency

tlater-famedly (changes requested):

.config/starte2e.sh
mzaniolo: If I don't restart the ldap service it don't get the new certificates. I think now it should be fine

.config/starte2e.sh
mzaniolo: What using an openssl container just for that? It very lightweight.

.config/starte2e.sh
tlater-famedly: ldap is also pretty slow to start; could we check the mtime on the certs or such to determine whether to restart?

.config/starte2e.sh
tlater-famedly: Yeah, that seems reasonable

.config/starte2e.sh
mzaniolo: Now I have the script checking the mtime and only generating it if it's older than 25 days. See if it's better.

tlater-famedly (approved):

.config/starte2e.sh
tlater-famedly: Yep, ta!

Issue #30, opened 2024-08-13
Allow non-SSO deployments


tlater-famedly:
As of 9995b4c SSO is mandatory, which is obviously not what we want.

The problem is that Zitadel currently sends confirmation emails if no authentication method is enabled on user creation - and since we don't want to pre-set passwords for security reasons, but still don't want to email users for flow reasons, this means we cannot currently provide sync for users that should not use SSO.

Fixing this requires upstream changes to Zitadel, so this is blocked until we have some kind of resolution for https://github.com/zitadel/zitadel/issues/8422

nikzen:
This is possible now:

POST https://auth.staging.famedly.de/v2/users/human

```
{
  "userId": "1234567abc",
  "username": "larodar1234567abc",
  "profile": {
    "givenName": "Minnie",
    "familyName": "Mouse",
    "displayName": "Minnie Mouse"
  },
  "email": {
    "email": "n.zender+larodar1234567abc@famedly.com",
    "isVerified": true
  }
}
```

Creates a user and **no** mail is sent. If the user tries to login, zitadel sends a code to set the password

Issue #24, opened 2024-08-09
Switch to new API


nikzen:
### Description
There is a new API -> https://zitadel.com/docs/apis/resources/user_service_v2/user-service-add-human-user. This API would allow us to combine the create user & add metadata into one api call. It would also remove the nickname workaround, as we will be able to generate our own userId.

### userId
- With the old API, the Zitadel userId was generated by Zitadel.
- With the new API, we are able to set our own userID.
- We need to find a rule of how to generate this userId.
- Suggestion: uuidv5-famedly-namespace (organizationId:idpId:user_id)

An example API call might look like this:

```
CURL
curl -L 'https://$CUSTOM-DOMAIN/v2/users/human' \
-H 'Content-Type: application/json' \
-H 'Accept: application/json' \
-H 'Authorization: Bearer <TOKEN>' \
--data-raw '{
  "userId": "generatedUserId",
  "username": "email",
  "profile": {
    "givenName": "first_name",
    "familyName": "last_name",
    "displayName": "first_name, last_name"
  },
  "email": {
    "email": "email",
    "isVerified": true
  },
  "phone": {
    "phone": "phone",
    "isVerified": true
  },
  "metadata": [
    {
      "key": "localpart",
      "value": "uuidv5(user_id)"
    }
  ],
  "idpLinks": [
    {
      "idpId": "idpId",
      "userId": "user_id",
      "userName": "preferred_username"
    }
  ]
}
```

Deactivate / delete 

Currently, we are using the nickname to search for users so that we can figure out the generated userId and then delete the user. We can remove the search API call, and directly delete all users based on the userId, as we know how the userId looks like. However, if we are in the deactivate-only mode we do not know the userid and have to search for the users, based on the mail address.

emgrav:
Looks like the Zitadel crate already has support for v2 APIs in versions >=4.0.0:

https://github.com/smartive/zitadel-rust/releases/tag/v4.0.0

Issue #75, opened 2024-10-18
Add a user ID column to the CSV import format


tlater-famedly:
To track changes to users between updates, we need to be able to uniquely identify them. Since the source is authoritative, an ID from the source must be used to do this.

Currently, the CSV source re-uses the user's email address as this ID, but since this is PII, this means we cannot pseudonymise CSV entries. We should add a new column to enable this.

Things to consider:

- This will be a *very* breaking change, and since the tool is already used in production here and there, we might need a way to migrate.
  - This could simply be a check for existing users with the given email address as an external ID, but it would be nice to find a way to prevent us having to lug around code for this forever...
- Technically the ID column can be entirely free-form, but we may want to enforce it containing UUIDs or simple numeric IDs to prevent user error causing PII leaks.

jannden:
After clarification with Niklas, this task talks about **external user IDs**. This is different from #96 , which talks about **zitadel user IDs**.

tlater-famedly:
They are in fact the same, as #96 is about using the "localpart" as the Zitadel ID, which is just the external user ID encoded with UUIDv5 in our namespace. You cannot get the original value from a UUIDv5, but we *could* have just had one field for the ID, and encoded it.

PR #79, opened 2024-10-30
Remove ldap caching and move to the Zitadel v2 API

tlater-famedly:
Closes: https://github.com/famedly/famedly-sync/issues/24
Closes: https://github.com/famedly/famedly-sync/issues/53
Closes: https://github.com/famedly/famedly-sync/issues/30

WIP because there are some things to clear up still. The tests will all fail here, but with some fudging of the sources *almost* all tests pass.

There are a few things that need to be discussed/covered:

### Multi-source config/test suite

Previously, we wanted to support specifying multiple sources simultaneously, since that came somewhat for free. With the removal of the cache, however, *only* the UKT source uses email addresses to identify users, and in fact only it has a concept of "removed" users, which means that it fundamentally needs to be handled separately from the other sources. Trying to support multi-source definitions is much more tricky with this limitation, so I believe we should not do so.

This is the cause of most of the remaining failures; the test suite was written around the assumption that we can just dump all kinds of sources into one big config file, and for getting this over the line I've ignored also rewriting the test suite so far, instead testing by only having the respective source actively usable in code for the time being.

### External user ID encoding

Since we universally use the external user IDs to uniquely identify users now, and we actually pull data from Zitadel instead of just pushing, we now need to re-ingest the user IDs we've written to the nickname field. Unfortunately, because of ldap3's odd behavior of silently converting byte-typed values to strings when they can be, but otherwise returning bytes, we need to base64-encode these IDs.

So far we've just encoded any values that were bytes, and left strings alone - alas, apparently you can end up with strings randomly being valid base64 values (e.g. `starttls`). This means that we do not have a way to distinguish between IDs that were encoded and IDs that were not - when we get unlucky, this can make it impossible to uniquely identify a user, since decoding stringly-typed IDs can sometimes result in valid values.

Fixing this in the current implementation is easy (simply prefix encoded strings with a prefix, or encode *all* values, since UTF-8 strings are also valid byte strings), but backwards compatibility is a problem, since users with encoded IDs already exist in production.

It's possible that there is no way around runs in which some users will be deleted and immediately re-created.

### Zitadel version

Due to the various recent Zitadel bugs, we need to update the Zitadel version in this PR. I've currently gone with the latest (`2.64.1`), but maybe we want another version?

---

Remaining tasks:

- [x] [Famedly-Sync: WIP Refactor getting rid of cache & single-source sync | famedly-sync#90](https://github.com/famedly/famedly-sync/issues/90)
- [x] [Support single-source sync in Tests | famedly-sync#81](https://github.com/famedly/famedly-sync/issues/81)
- [x] [Famedly-Sync: Refactor for ID encoding and sorting | famedly-sync#91](https://github.com/famedly/famedly-sync/issues/91)
- [x] [Debug and fix tests for Famedly Sync after ID encoding changes | famedly-sync#86](https://github.com/famedly/famedly-sync/issues/86)
- [x] [Encoding library that preserves alphabetical order | famedly-sync#87](https://github.com/famedly/famedly-sync/issues/87)
- [ ] [Migration executable for encoding old IDs | famedly-sync#88](https://github.com/famedly/famedly-sync/issues/88)
- [ ] [Documentation for changes and migration guide | famedly-sync#89](https://github.com/famedly/famedly-sync/issues/89)


emgrav:
> It's possible that there is no way around runs in which some users will be deleted and immediately re-created.

For UKE we'll need to make sure to test this with a dry run

src/lib.rs
```diff
@@ -1,12 +1,253 @@
...
+			.context(format!("Failed to query users from {}", source.get_name()))
+	}
+
+	let csv = config.sources.csv.clone().map(CsvSource::new);
+	let ldap = config.sources.ldap.clone().map(LdapSource::new);
+	let ukt = config.sources.ukt.clone().map(UktSource::new);
+
+	// The ukt source is handled specially, since it doesn't behave as
+	// the others
+	if let Some(ukt) = ukt {
```
tlater-famedly, line 63: This doesn't correctly check that exactly *one* source is defined, which is unfortunate. Couldn't think of a good way to express this condition, while also treating ukt specially, but DRY-ing out the other sources.

src/zitadel.rs
```diff
@@ -41,434 +58,263 @@ impl Zitadel {
...
-					None,
-					None,
-				)
-				.await?;
-		};
+		if self.feature_flags.is_enabled(FeatureFlag::SsoLogin) {
+			user.set_idp_links(vec![IdpLink::new()
+				.with_user_id(imported_user.external_user_id.to_string())
+				.with_idp_id(self.zitadel_config.idp_id.clone())
+				// TODO: Figure out if this is the correct value; empty is not permitted
```
tlater-famedly, line 178: This should probably be confirmed to work before we merge; we have no way to test if IDP logins *actually* work.

tlater-famedly:
> > It's possible that there is no way around runs in which some users will be deleted and immediately re-created.
> 
> For UKE we'll need to make sure to test this with a dry run

The UKE source *should* be unaffected, because it doesn't need to parse external IDs from Zitadel. But yeah, probably prudent to do dry runs for the first sync after updating in general, this is a pretty big change.

emgrav (approved):

emgrav:
Remaining before this can be merged:

1. Some tests need to be refactored to no longer combine multiple sources.
2. We need to document the potential quirks of external user ID encoding, including instructions for infra on what to look for in a dry run to identify a problem before it happens.

jannden:
There is a clash between CSV and LDAP at this point. When we import CSV, it deletes all Zitadel users not present in the CSV.
A) Is that a desired behavior?
B) The old and new users are compared with their `external_user_id` which won't match if we mix CSV and LDAP. But CSV and LDAP use different `external_user_id`. CSV uses email addresses for that and LDAP uses user_id attribute pulled from LDAP for that. Shall we add another field in CSV to explicitly add a `user_id` that has the potential to match LDAP `user_id`?

tlater-famedly:
> There is a clash between CSV and LDAP at this point. When we import CSV, it wrongly deletes Zitadel users not present in the CSV (or rather users that it couldn't match).

> A) Is that a desired behavior?

Ideal behavior would of course be that switching between LDAP/CSV is possible, but this is not a use case we actually currently have, so while not ideal, it's probbaly *fine* for this to happen.

>  B) The old and new users are compared with their external_user_id which won't match if we mix CSV and LDAP, because they use different external_user_id. CSV uses email addresses for that and LDAP uses user_id attribute pulled from LDAP for that. Shall we add another field in CSV to explicitly add a user_id that has the potential to match LDAP user_id?

Yes, this is already a planned feature, see #75. We should do that as a separate task, though, as mixing the sources will for now be unsupported.

We should probably make sure with product that this is indeed fine, however.

jannden:
Okay, in that case, we can merge the updated tests from: https://github.com/famedly/famedly-sync/pull/80

emgrav (dismissed):

jannden (approved):

emgrav:
When we merge this we should tag a new version and notify Niklas

src/user.rs
```diff
@@ -20,107 +19,102 @@ pub(crate) struct User {
...
+			.and_then(|human_email| human_email.email())
+			.ok_or(anyhow!("Missing email address for {}", external_id))?
+			.clone();
 
-	/// Return the name to be used in logs to identify this user
-	pub(crate) fn log_name(&self) -> String {
-		format!("email={}", &self.user_data.email)
+		let phone = user.phone().and_then(|human_phone| human_phone.phone());
+
+		let external_user_id = match BASE64_STANDARD.decode(external_id.clone()) {
```
tlater-famedly, line 50: I've had to think really hard about how this all works while trying to write a migration guide for this, and realized that the current implementation simply cannot work for sources that have non-UTF8 external user IDs (read: AD).

The issue is that base64 encoding values **apparently does *not* preserve alphabetic ordering**. This is an issue, because it means that the order our source returns will be different from the one Zitadel returns if Zitadel deals with base64-encoded values and the source does not, thereby completely breaking the sync algorithm.

It's unfortunate that I realized this this far down the line...

We will need to handle this completely differently. I'd propose already encoding binary IDs at the source level (i.e., when parsing IDs from *any* source), and only ever handling encoded values for IDs elsewhere in the codebase. This guarantees that the sort order is consistent. We should also never pass `ByteOrString` enums into the public interface of the `user` module - Zitadel should only ever deal with `String`s anyway, so this will clean up the codebase, too.

This will also mean that Zitadel will only ever see encoded values for IDs, making the decode step completely unnecessary.

In addition, since base64 would prevent us from ever using source APIs to sort incoming users instead of doing it on the application end, incurring pointless memory overhead, we should not use base64 encoding to do this, but instead an encoding scheme that *does* preserve alphabetical order. I'm not aware of any off the top of my head, but I am sure they exist.

This also gives us a clean migration path, as we can write a separate command that goes through all users of a Zitadel instance and re-encodes their IDs in that encoding, which must then be run on any Zitadel instances that have previously been synced to using this tool.

On the up side, since the ldap attribute parsing implementation [was much more conscientious than I remembered](https://github.com/famedly/famedly-sync/blob/6f6db25e0d2688beed547b6c5605e3df987ce3c6/src/sources/ldap.rs#L112), there should be no instances with mixed binary/string IDs, so such a tool can have subcommands for binary/string mode re-encoding, and we should not run any particular risk of accidentally decoding a value that was originally intended to be a string.

Unfortunately, Zitadel [still requires this broken way of passing through user IDs](https://github.com/zitadel/zitadel/blob/main/internal/idp/providers/ldap/session.go#L289) for IDP IDs, so we will need to decode the value we just encoded just to re-encode it with a different encoding scheme if the original value happens to not be valid UTF8. At least it helps me with my impostor syndrome to know that other programmers clearly stumble over the same quirks of ldap.

Sorry for the delay in all this, this has been *so* much more complex than it seems at face value.

src/zitadel.rs
tlater-famedly: Since I ended up reading the source code to figure out how Zitadel handles IDP IDs, I also figured out what the "name" means. I originally just wrote this comment:

> This refers to the "username" that the search filter defined in the IDP config resolves to. This means that for our IDP links to work, that needs to be configured to resolve to the user's email address. See: https://github.com/zitadel/zitadel/blob/250f2344c8c2292ca9b861cdd12223d0b4719d43/internal/idp/providers/ldap/session.go#L230

That should make it into the readme, though.

src/user.rs
emgrav: Thank you for the extensive problem description. Let's talk about it in our sync tomorrow.

Regarding the encoding scheme, I think base32hex is a good option here: https://www.rfc-editor.org/rfc/rfc4648#page-10

tests/e2e.rs
```diff
@@ -22,15 +22,208 @@ use zitadel_rust_client::v1::{
 	Email, Gender, ImportHumanUserRequest, Phone, Profile, UserType, Zitadel,
 };
```
jannden, line 23: The v1 & v2 use should be updated/verified to match Zitadel operations from `src/zitadel.rs`

jannden (changes requested):

src/zitadel.rs
```diff
@@ -41,434 +48,256 @@ impl Zitadel {
...
-					None,
-					Some(new.get_display_name()),
-					None,
-					None,
-				)
-				.await?;
-		};
+		if self.feature_flags.is_enabled(FeatureFlag::SsoLogin) {
+			user.set_idp_links(vec![IdpLink::new()
+				.with_user_id(imported_user.external_user_id.clone())
```
tlater-famedly, line 159: We need to use the broken base64-encoded scheme here for Zitadel to interpret the IDs "correctly".

We should probably also send an issue upstream about this?

tests/e2e.rs
```diff
@@ -836,18 +1115,21 @@ async fn test_e2e_ukt_sync() {
...
-	config.perform_sync().await.expect("syncing failed");
+	perform_sync(&config).await.expect("syncing failed");
 
 	let user = zitadel.get_user_by_login_name("delete_me@famedly.de").await;
 	assert!(user.is_err_and(|error| matches!(error, ZitadelError::TonicResponseError(status) if status.code() == TonicErrorCode::NotFound)));
 }
 
+// Currently fails because CSV uses non-hex encoded IDs, need to think
+// about how to fit this into the overall workflow
+#[ignore]
```
tlater-famedly, line 1126: Not working *yet*

src/zitadel.rs
tlater-famedly: Need a test to confirm this is correctly encoding IDs too.

tests/e2e.rs
tlater-famedly: Yeah, but this isn't particularly high priority. We only *need* to do this once/if the v1 API is deprecated, which doesn't look like it will be happening anytime soon; too many endpoints are still exclusive to v1 for that to be realistic.

So I suggest we put this on the pile of tech debt issues. We should clean up the test suite anyway.

tests/e2e.rs
```diff
@@ -1,17 +1,17 @@
 #![cfg(test)]
```
tlater-famedly, line 1: We should add one more test, to assert that SSO linking actually works. I *believe* Zitadel is set up to allow this already, we just need to add a test.

tests/e2e.rs
```diff
@@ -1,17 +1,17 @@
...
+#![allow(clippy::expect_fun_call)]
```
tlater-famedly, line 2: Should add this to the clippy config, and add it to our company-wide clippy config, if we don't like that lint.

tests/e2e.rs
tlater-famedly: I suppose the CSV backend should simply also hex-encode any IDs, and we simply make this an assumption of all sources so that we can rely on this being correct all the time?

tests/e2e.rs
jannden: Yeah, fair enough.

tests/e2e.rs
jannden: Yes, all IDs should go through the encoding

src/sources/csv.rs
```diff
@@ -76,12 +74,12 @@ impl CsvData {
...
-			last_name: csv_data.last_name.into(),
-			phone: if csv_data.phone.is_empty() { None } else { Some(csv_data.phone.into()) },
-			preferred_username: csv_data.email.clone().into(),
-			external_user_id: csv_data.email.into(),
+			email: csv_data.email.clone(),
+			first_name: csv_data.first_name,
+			last_name: csv_data.last_name,
+			phone: if csv_data.phone.is_empty() { None } else { Some(csv_data.phone) },
+			preferred_username: Some(csv_data.email.clone()),
+			external_user_id: csv_data.email,
```
jannden, line 82: So this should be hex encoded

src/user.rs
```diff
@@ -1,148 +1,127 @@
...
+	///
+	/// Only use this for Zitadel support.
+	pub fn get_string_id(&self) -> Result<String> {
+		let id = self.get_external_id_bytes()?;
+		Ok(if let Ok(encoded_id) = String::from_utf8(id.clone()) {
+			encoded_id
+		} else {
+			BASE64_STANDARD.encode(id)
+		})
+	}
```
jannden, line 100: This seems to be forgotten for the HEX encoding update.

src/user.rs
```diff
@@ -1,148 +1,127 @@
...
-				external_user_id: self.user_data.external_user_id.clone().to_string(),
-				display_name: self.get_display_name(),
-			}]
-		} else {
-			vec![]
-		}
+	/// Get a display name for this user
+	pub fn get_display_name(&self) -> String {
+		format!("{}, {}", self.last_name, self.first_name)
 	}
```
jannden, line 66: Are we actually disregarding the `display_name` from LDAP on purpose?
For example, in tests, when we do:

```rust
	let config = ldap_config().await;
	let mut ldap = Ldap::new().await;
	ldap.create_user(
		"John",
		"Doe",
		"WHATEVER DISPLAY NAME", // Whatever we put here, it won't be picked up by the sync
		email,
		Some(phone_number),
		uid,
		enabled,
	)
	.await;

	perform_sync(config).await.expect("LDAP sync failed");
```

The display name will be ignored.

jannden (changes requested):
A couple of things I noticed, meaning as points of discussion.

src/main.rs
```diff
@@ -20,7 +20,7 @@ async fn main() -> ExitCode {
 #[allow(clippy::print_stderr)]
 async fn run_sync() -> Result<()> {
 	let config = {
-		let config_path = std::env::var("FAMEDLY_SYNC_CONFIG").unwrap_or("config.yaml".into());
+		let config_path = std::env::var("FAMEDLY_LDAP_SYNC_CONFIG").unwrap_or("config.yaml".into());
```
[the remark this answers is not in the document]
tlater-famedly: I think the variable is just misnamed?

src/user.rs
tlater-famedly: No, give the doc comment another read. Zitadel uses base64 encoding to map binary IDP IDs, and will not understand our hex-encoded IDs. We need to encode IDs as base64 *specifically* for the cases in which zitadel actually reads them.

Maybe we should move this to `zitadel.rs` though?

src/user.rs
tlater-famedly: Yes, this is as per the design: https://www.notion.so/famedly/Famedly-Sync-Agent-a7e1167785d44393a29083e81bc43bf4?pvs=4#c33cbdabda1144659da80ce2e41596b1

src/user.rs
jannden: Got it. I will move it to `zitadel.rs`

tests/e2e.rs
jannden: I suggest to create an issue out of this and leave it explicitly in the test file for this PR.

jannden (dismissed):
Good for manual tests now.

src/zitadel.rs
tlater-famedly: This is complex to write since openLDAP does not support binary IDs.

We should probably reconsider how the test suite handles IDs together with #97

tests/e2e.rs
tlater-famedly: This has a (small) performance impact. While this isn't important in tests, there's no way to tell clippy that we don't care about it in tests specifically, so I just fixed it instead.

We can raise an issue to discuss whether we want to trade-off this bit of performance for legibility separately.

jannden (approved):

PR #80, opened 2024-11-04
test: Support single-source sync

jannden:
Closes: https://github.com/famedly/famedly-sync/issues/81

This fixes tests for: https://github.com/famedly/famedly-sync/pull/79

As per this [comment](https://github.com/famedly/famedly-sync/pull/79#issuecomment-2449942466).

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -19,6 +19,7 @@ feature_flags:
   # - deactivate_only # Only deactivate users, do not create or update them. Keep in mind LDAP is cached and all the changes made on LDAP will be written to the cache as if they where applied. Therefore, only the deactivation changes will be applied to Zitadel but **all the other changes will be lost**.
 
 # Configuration for the sources to sync from.
+# IMPORTANT: Only one source can be present at a time.
 sources:
   # Configuration for the LDAP source. Using caching, LDAP source checks for new, updated, and deleted users in the LDAP server.
   ldap:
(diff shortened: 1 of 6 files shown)


jannden:
As mentioned in the original PR [in this comment](https://github.com/famedly/famedly-sync/pull/79#issuecomment-2453335474), we first need to clarify how to approach the clash between LDAP and CSV sync. Then a small adjustment to the `test_e2e_full_sync` is expected and this should be ready to go.

emgrav (approved):

config.sample.yaml
```diff
@@ -19,6 +19,7 @@ feature_flags:
   # - deactivate_only # Only deactivate users, do not create or update them. Keep in mind LDAP is cached and all the changes made on LDAP will be written to the cache as if they where applied. Therefore, only the deactivation changes will be applied to Zitadel but **all the other changes will be lost**.
 
 # Configuration for the sources to sync from.
+# IMPORTANT: Only one source can be present at a time.
```
tlater-famedly, line 22: I think the sample config should be split in three for this, don't want to document something that can never work.

tlater-famedly (commented):
Looks good, I like this!

config.sample.yaml
jannden: Okay, let's split the sample as part of the documentation commit on the feature branch.

Issue #86, opened 2024-11-08
Debug and fix tests for Famedly Sync after ID encoding changes


jannden:
More info [here](https://github.com/famedly/famedly-sync/pull/79#discussion_r1830209026).

Issue #87, opened 2024-11-08
Encoding library that preserves alphabetical order


jannden:
Investigate and find an encoding library that preserves alphabetical order.
If such library was found, implement it.
Start with:  https://github.com/hsivonen/encoding_rs

From [here](https://github.com/famedly/famedly-sync/pull/79#discussion_r1830209026):
> Since base64 would prevent us from ever using source APIs to sort incoming users instead of doing it on the application end, incurring pointless memory overhead, we should not use base64 encoding to do this, but instead an encoding scheme that does preserve alphabetical order. I'm not aware of any off the top of my head, but I am sure they exist. In the short term this is not essential, as we could migrate to a different encoding in the future, but we might as well pick a future proof encoding at this point.

emgrav:
> encoding_rs

Is this the right crate for this? I think we want to use something like `base32hex`, one crate that provides an implementation of this is https://docs.rs/data-encoding/latest/data_encoding/index.html

jannden:
@emgrav Yup, I believe you are pointing to the right direction.
I am wondering though why not to use just a simple `hex` encoding where we won't have to worry about padding.
I will create a WIP and do some tests to see how it works.

PR #92, opened 2024-11-11
refactor: External ID encoding with lexicographical order

jannden:
Closes: #86 & #87

### DIFF
--- a/src/sources/ldap.rs
+++ b/src/sources/ldap.rs
@@ -4,7 +4,6 @@ use std::{fmt::Display, path::PathBuf};
 
 use anyhow::{anyhow, Context, Result};
 use async_trait::async_trait;
-use base64::prelude::{Engine, BASE64_STANDARD};
 use ldap_poller::{
 	config::TLSConfig, ldap::EntryStatus, ldap3::SearchEntry, AttributeConfig, CacheMethod,
 	ConnectionConfig, Ldap, SearchEntryExt, Searches,
@@ -90,9 +89,9 @@ impl LdapSource {
 			.any(|flag| status_as_int & flag != 0);
 
 		let ldap_user_id = match read_search_entry(&entry, &self.ldap_config.attributes.user_id)? {
-			// TODO(tlater): Use an encoding that preserves alphabetic order
-			StringOrBytes::Bytes(byte_id) => BASE64_STANDARD.encode(byte_id),
-			StringOrBytes::String(string_id) => BASE64_STANDARD.encode(string_id),
+			// Use hex encoding instead of base64 for consistent alphabetical order
+			StringOrBytes::Bytes(byte_id) => hex::encode(byte_id),
+			StringOrBytes::String(string_id) => hex::encode(string_id.as_bytes()),
 		};
 
 		let first_name =
@@ -128,9 +127,11 @@ fn read_string_entry(
 ) -> Result<String> {
 	match read_search_entry(entry, attribute)? {
 		StringOrBytes::String(entry) => Ok(entry),
-		StringOrBytes::Bytes(_) => {
-			Err(anyhow!("Unacceptable binary value for {} of user `{}`", attribute, id))
-		}
+		StringOrBytes::Bytes(_) => Err(anyhow!(
+			"Binary values are not accepted: attribute `{}` of user `{}`",
+			attribute,
+			id
+		)),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


src/sources/ldap.rs
```diff
@@ -124,13 +124,11 @@ impl LdapSource {
...
 	attribute: &AttributeMapping,
-	id: &str,
+	_id: &str,
 ) -> Result<String> {
 	match read_search_entry(entry, attribute)? {
 		StringOrBytes::String(entry) => Ok(entry),
-		StringOrBytes::Bytes(_) => {
-			Err(anyhow!("Unacceptable binary value for {} of user `{}`", attribute, id))
-		}
+		StringOrBytes::Bytes(bytes) => Ok(BASE64_STANDARD.encode(&bytes)),
```
jannden, line 131: @tlater-famedly
I couldn't immediately understand why wasn't Bytes accepted?
Also, shall Base64 be sufficient here?

src/sources/ldap.rs
```diff
@@ -90,9 +90,9 @@ impl LdapSource {
 			.any(|flag| status_as_int & flag != 0);
 
 		let ldap_user_id = match read_search_entry(&entry, &self.ldap_config.attributes.user_id)? {
-			// TODO(tlater): Use an encoding that preserves alphabetic order
-			StringOrBytes::Bytes(byte_id) => BASE64_STANDARD.encode(byte_id),
-			StringOrBytes::String(string_id) => BASE64_STANDARD.encode(string_id),
+			// Use hex encoding instead of base64 for consistent alphabetical order
+			StringOrBytes::Bytes(byte_id) => hex::encode(byte_id),
+			StringOrBytes::String(string_id) => hex::encode(string_id.to_lowercase().as_bytes()),
```
jannden, line 95: @emgrav You suggested `base32hex`, was it for a particular reason? I think just a plain `hex` encoding should be sufficient without having to worry about padding.

emgrav: Yes, regular hex also works. base32hex is slightly more space efficient but that's of no concern here.

src/sources/ldap.rs
tlater-famedly: This function's entire point was throwing an error for non-string values, it's only here to not write the same error handling 6 times x) Hence the doc comment:

```rust
/// Read an an attribute, but assert that it is a string
```

I don't think base64 encoding these values is a good idea, it will be stuff like user names, which should always be strings. We probably should skip syncing users which have non-string-encoded values that we expect to be strings.

This is different for the values that we expect to be raw binary, for those we use the `read_search_entry` function, and handle the non-string case explicitly.

This *is* indeed a change, but I think this just makes more sense. Happy to hear counter arguments.

src/sources/ldap.rs
tlater-famedly, line 95: Are (open)ldap IDs not case sensitive?

tests/e2e.rs
```diff
@@ -485,131 +682,135 @@ async fn test_e2e_sync_invalid_phone() {
...
-// 			.expect("could not get user metadata");
-
-// 		assert_eq!(
-// 			preferred_username
-// 				.map(|u| BASE64_STANDARD.decode(u).expect("failed to decode binary attr")),
-// 			Some("validutf8".as_bytes().to_vec())
-// 		);
-// 	}
-// }
+#[test(tokio::test)]
```
tlater-famedly, line 685: Yep, hence these tests were commented out; I think we should not support this use case anymore.

We *should* test that binary user IDs in fact work. Maybe we can change the attribute we use for the ID to `userSMIMECertificate` with the config?

jannden: I have created a new test and updated the two old.

src/sources/ldap.rs
jannden: Okay, reverted.

src/sources/ldap.rs
jannden: Not sure about this. Let's discuss.

tlater-famedly: Well, if they are case sensitive this will cause collisions, while if it they are case sensitive there's a chance that they might change the casing randomly.

I assume the latter is probably not the case, and we should just remove the `to_lowercase()`. Is there a particular reason we do that?

jannden: Since there is not a particular reason, I agree with you - it's logical to remove the lowercase conversion.

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
+//! This binary is used to migrate user IDs from base64 to hex encoding.
```
tlater-famedly, line 1: Can we do this in a separate PR so we get the base functionality ready for a test run quicker?

tlater-famedly (approved):

PR #93, opened 2024-11-14
feat: Migration script for LDAP sync

### LINKED ISSUE #88: Migration executable for encoding old IDs
Write migration command for deployed instances of ldap-sync
- in Zitadel, currently the ids are either not encoded or base64 encoded, so we need to go through every instance and appropriately re-encode it in the new way
- this should be created as a secondary entry point with different binary within ldap-sync

More info [here](https://github.com/famedly/famedly-sync/pull/79#discussion_r1830209026).

jannden:
Closes #88

### DIFF
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -90,16 +90,22 @@ fn detect_database_encoding(users: Vec<SyncUser>) -> ExternalIdEncoding {
 		}
 	}
 
+	// Return early if no valid samples
+	if total == 0 {
+		return ExternalIdEncoding::Ambiguous;
+	}
+
 	// Use thresholds to determine encoding
 	let hex_ratio = f64::from(hex_count) / f64::from(total);
 	let base64_ratio = f64::from(base64_count) / f64::from(total);
 
-	if hex_ratio > 0.8 {
-		ExternalIdEncoding::Hex
-	} else if base64_ratio > 0.8 {
-		ExternalIdEncoding::Base64
-	} else {
-		ExternalIdEncoding::Plain
+	// Require a strong majority (90%) for a format to be considered dominant
+	// Also detect when both formats have significant presence
+	match (hex_ratio, base64_ratio) {
+		(h, _) if h > 0.9 => ExternalIdEncoding::Hex,
+		(_, b) if b > 0.9 => ExternalIdEncoding::Base64,
+		(h, b) if h > 0.2 && b > 0.2 => ExternalIdEncoding::Ambiguous, // Both formats present
+		_ => ExternalIdEncoding::Ambiguous,                            // No clear dominant format
 	}
 }
 
@@ -185,7 +191,7 @@ mod tests {
 		// "cafe" might count for both hex and base64, but "cafeb" and "abc" won't count
 		// for either. Out of 3, maybe 1 counts as hex/base64 and 2 are plain. Ratios:
 		// hex = 1/3 ≈ 0.33, base64 = 1/3 ≈ 0.33, both < 0.8.
-		run_detection_test(user_ids, ExternalIdEncoding::Plain);
+		run_detection_test(user_ids, ExternalIdEncoding::Ambiguous);
 	}
 
 	#[tokio::test]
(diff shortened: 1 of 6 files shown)


src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+	tracing::info!("Starting migration");
+	tracing::debug!("Old external IDs will be base64 decoded and re-encoded as hex");
+	tracing::debug!("Note: External IDs are stored in the nick_name field of the user's profile in Zitadel, often referred to as uid.");
+
+	// Zitadel
+	let zitadel_config = config.zitadel.clone();
+	let mut zitadel = Zitadel::new(zitadel_config.url, zitadel_config.key_file).await?;
+
+	// Get all users
+	let mut stream = zitadel.list_users(
```
tlater-famedly, line 45: Should we reuse the method from the ldap source for this? Then we don't have to do the user conversion here as well.

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+		ListUsersRequest::new(vec![
+			SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human))
+		])
+		.with_asc(true)
+		.with_sorting_column(UserFieldName::NickName),
+	)?;
+
+	// Process each user
+	while let Some(user) = stream.next().await {
+		let user_id = user.user_id().ok_or_else(|| anyhow!("User lacks ID"))?;
```
tlater-famedly, line 55: Even if we don't reuse the method for getting the stream, we should probably reuse the user parsing as much as possible.

I wouldn't suggest doing so if we need to add options to the type to avoid parsing specific fields, of course, I'm not in the codebase enough right now to spot how awkward this would be with the existing types, and how much they could be redesigned to fit this purpose as well.

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+		let given_name = profile.given_name();
+		let family_name = profile.family_name();
+		let display_name = profile.display_name();
+
+		let Some(nick_name) = profile.nick_name() else {
+			tracing::warn!(user_id = %user_id, "User lacks nick_name (external uid), skipping");
+			continue;
+		};
+
+		tracing::info!(user_id = %user_id, old_uid = %nick_name, "Starting migration for user");
```
tlater-famedly, line 68: This is fine, but the user struct deliberately implements display to prevent accidental future leaks of PII in logs. If we don't use the user struct, we should maybe leave a comment here to make sure nobody is tempted to start printing email addresses in the future.

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+
+		let new_external_id = if nick_name.is_empty() {
+			// Keep empty string as is, will skip it later
```
tlater-famedly, line 71: Why do we not just end early here? There are three cases, this would be cleaner as a `match`.

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+			nick_name.to_string()
+		} else {
+			// First check if it's valid hex
+			if nick_name.chars().all(|c| c.is_ascii_hexdigit()) && nick_name.len() % 2 == 0 {
+				// If valid hex, keep as is
+				tracing::debug!(  user_id = %user_id, old_uid = %nick_name,"Valid hex uid detected, keeping as is");
```
tlater-famedly, line 77: The valid set of characters for a hex-encoded string is a subset of the ones for a base64 encoded string. As such, while unlikely, this could end up mis-identifying some base64-encoded IDs as hex encoded.

We can still do this kind of check, but we should probably employ it as a heuristic (that can be overridden) on the whole database instead of per-record.

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+				nick_name.to_string()
+			} else {
+				// Try base64, if fails use plain text
```
tlater-famedly, line 80: Hmm, IDs which happen to be valid base64 will be "decoded" in this case. I realize we're working with spotty data here, because of how we encoded the IDs in the first place, but is there anything we can do to improve this?

Notably, due to the original implementation, the database should be either fully base64-encoded, or fully string-encoded. Maybe again a full-database heuristic to reduce the chances of failure here?

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+        new_uid = %new_external_id,
+        "Skipping user due to empty uid");
+			continue;
+		}
+
+		// Update uid (external ID, nick_name) in Zitadel
+		let mut request = UpdateHumanUserRequest::new();
+		request.set_profile(
+			SetHumanProfile::new(
+				given_name.ok_or_else(|| anyhow!("User lacks given name"))?.to_owned(),
```
tlater-famedly, line 106: Are these mandatory? Can't we just update the ID?

tests/e2e.rs
```diff
@@ -1396,6 +1397,112 @@ async fn test_e2e_sso_linking() {
 	);
 }
 
+#[test(tokio::test)]
+#[test_log(default_log_filter = "debug")]
+async fn test_e2e_migrate_base64_id() {
```
tlater-famedly, line 1402: Can these tests be put in a separate file?

tests/e2e.rs
```diff
@@ -1396,6 +1397,112 @@ async fn test_e2e_sso_linking() {
 	);
 }
```
tlater-famedly, line 1399: We should add tests for the edge cases I mention further up; IDs that happen to be valid base64/hex, but aren't intended to be treated as such originally.

tlater-famedly (changes requested):

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+				family_name.ok_or_else(|| anyhow!("User lacks family name"))?.to_owned(),
+			)
+			.with_display_name(
+				display_name.ok_or_else(|| anyhow!("User lacks display name"))?.to_owned(),
+			)
+			.with_nick_name(new_external_id.clone()),
+		);
+
+		zitadel.update_human_user(user_id, request).await?;
```
tlater-famedly, line 115: We should respect the `DryRun` flag.

src/bin/migrate.rs
jannden: Which method do you mean? `get_next_zitadel_user`?

tlater-famedly: This method: https://github.com/famedly/famedly-sync/blob/main/src/zitadel.rs#L91

src/bin/migrate.rs
```diff
@@ -0,0 +1,127 @@
...
+			}
+		};
+
+		// Skip empty uid
+		if new_external_id.is_empty() {
+			tracing::warn!(
+        user_id = %user_id,
+        old_uid = %nick_name,
+        new_uid = %new_external_id,
+        "Skipping user due to empty uid");
```
sirewix, line 98: broken indentation

src/zitadel.rs
```diff
@@ -327,6 +310,24 @@ fn search_result_to_user(user: ZitadelUser) -> Result<User> {
...
+/// create collisions (although this is unlikely).
+///
+/// Only use this for Zitadel support.
+pub fn get_zitadel_encoded_id(external_id_bytes: Vec<u8>) -> Result<String> {
+	Ok(if let Ok(encoded_id) = String::from_utf8(external_id_bytes.clone()) {
+		encoded_id
+	} else {
+		BASE64_STANDARD.encode(external_id_bytes)
+	})
+}
```
sirewix, line 329:
```suggestion
pub fn get_zitadel_encoded_id(external_id_bytes: Vec<u8>) -> String {
	String::from_utf8(external_id_bytes.clone().unwrap_or_else(||
		BASE64_STANDARD.encode(external_id_bytes)
	)
}
```

src/bin/migrate.rs
jannden: Took a bit of refactoring, but here we go.

src/zitadel.rs
```diff
@@ -20,16 +20,20 @@ use zitadel_rust_client::{
...
+	get_next_zitadel_user,
 	user::User,
 	FeatureFlag,
 };
 
 /// The Zitadel project role to assign to users.
 const FAMEDLY_USER_ROLE: &str = "User";
 
+/// The number of users to sample for encoding detection
+const USER_SAMPLE_SIZE: usize = 10;
```
skomski, line 32: The sample size should be higher like 50+.

src/bin/migrate.rs
```diff
@@ -0,0 +1,257 @@
...
+	// Use thresholds to determine encoding
+	let hex_ratio = f64::from(hex_count) / f64::from(total);
+	let base64_ratio = f64::from(base64_count) / f64::from(total);
+
+	if hex_ratio > 0.8 {
+		ExternalIdEncoding::Hex
+	} else if base64_ratio > 0.8 {
+		ExternalIdEncoding::Base64
+	} else {
+		ExternalIdEncoding::Plain
```
skomski, line 102: This function should be stricter. The migration tool should not continue if one format is not dominant and Plain should not be the default option when the others are not dominant. There should be like an `ambiguous` option.
Aside from that what happens if the migration fails in-between for whatever reason? Then it should not be possible to restart the migration tool because it could lead to a wrong outcome. First the database should be rolled-back, or?

src/bin/migrate.rs
```diff
@@ -0,0 +1,257 @@
...
+	tracing::info!("Starting migration");
+	tracing::debug!("Old external IDs will be base64 decoded and re-encoded as hex");
+	tracing::debug!("Note: External IDs are stored in the nick_name field of the user's profile in Zitadel, often referred to as uid.");
+
+	// Zitadel
+	let mut zitadel = SyncZitadel::new(&config).await?;
+
+	// Detect external ID encoding based on a sample of users
+	let users_sample = zitadel.get_users_sample().await?;
+	let encoding = detect_database_encoding(users_sample);
```
skomski, line 41: Why does the tool continue when the dominant encoding is `hex`? Right now it would not convert the possible remaining base64 user_ids to hex because it would expect hex in `create_user_with_converted_external_id`, or?

skomski (changes requested):
Added some questions that I encountered.

src/bin/migrate.rs
jannden: Regarding the first point:
- Now an `ambiguous` option is introduced and `plain` is not default; when `ambiguous` is detected from the database heuristics, then it will go to check the particular case again regardless of the rest of the database.
- Fair point about the migration failing in-between, then a re-run would be tricky. After talking with @emgrav, for now it's acceptable to require any migration run to be done under controlled circumstances, i.e. with a db dump of zitadel taken right before, so we can just roll back; and in the long term, we should make it transactional.

src/bin/migrate.rs
jannden: The tool continues so that we can get logs about the particular cases, ie. we will see how many users don't match the db-detected encoding. Yes, it would not convert the possible `base64` user IDs to `hex` if `hex` was detected in DB heuristics. That's by design. Any better solution? Keep in mind that the result of a `hex` encoding is a subset of characters of the `base64` encoding. That's what prompted us with @tlater-famedly to implement DB heuristics.

skomski: Yes makes sense. Just wondered why it would not stop because there is no need to migrate.

src/bin/migrate.rs
skomski: Okay sounds good.

skomski (approved):

PR #95, opened 2024-11-15
Addressing PR comments

jannden:


src/main.rs
```diff
@@ -20,7 +20,7 @@ async fn main() -> ExitCode {
 #[allow(clippy::print_stderr)]
 async fn run_sync() -> Result<()> {
 	let config = {
-		let config_path = std::env::var("FAMEDLY_LDAP_SYNC_CONFIG").unwrap_or("config.yaml".into());
+		let config_path = std::env::var("FAMEDLY_SYNC_CONFIG").unwrap_or("config.yaml".into());
```
tlater-famedly, line 23: This is a breaking change and the commit at least needs to be marked as such. Not 100% sure it's worth doing.

src/zitadel.rs
```diff
@@ -306,6 +310,24 @@ fn search_result_to_user(user: ZitadelUser) -> Result<User> {
...
+/// or a UTF-8 string if not.
+///
+/// Note: This encoding scheme is inherently broken, because it is
+/// impossible to tell apart base64 encoded strings from
+/// non-base64 encoded strings. We can therefore never know if the
+/// ID should be decoded or not when re-parsing it, and it may
+/// create collisions (although this is unlikely).
+///
+/// Only use this for Zitadel support.
+pub fn get_string_id(external_id_bytes: Vec<u8>) -> Result<String> {
```
tlater-famedly, line 323:
```suggestion
pub fn get_zitadel_encoded_id(external_id_bytes: Vec<u8>) -> Result<String> {
```

tests/e2e.rs
```diff
@@ -1351,6 +1351,49 @@ async fn test_e2e_ldap_with_ukt_sync() {
...
+	assert!(!idps.is_empty(), "User should have IDP links");
+
+	let idp = idps.first().expect("No IDP link found");
+	assert_eq!(idp.idp_id, config.zitadel.idp_id, "IDP link should match configured IDP");
+	assert_eq!(idp.provided_user_id, test_uid, "IDP provided_user_id should match plain LDAP uid");
+	assert_eq!(idp.user_id, user.id, "IDP user_id should match Zitadel user id");
+	assert_eq!(
+		idp.provided_user_name, test_email,
+		"IDP provided_user_name should match test_email"
+	);
```
tlater-famedly, line 1394: Good, but perhaps we could add something to assert that SSO login *actually* works, by trying to log in as the user?

Can be a separate PR, but would be nice to confirm before we release a new version.

src/main.rs
tlater-famedly: (Also, we can address this in a branch/issue separate from the v2 api/cache-removal branch, so maybe we shouldn't do this here)

jannden: Wait, I believe this is just a revert of some mistaken rebase. We have FAMEDLY_SYNC_CONFIG on main:
https://github.com/famedly/famedly-sync/blob/main/src/main.rs#L23

src/sources/csv.rs
```diff
@@ -79,7 +79,7 @@ impl CsvData {
 			last_name: csv_data.last_name,
 			phone: if csv_data.phone.is_empty() { None } else { Some(csv_data.phone) },
 			preferred_username: Some(csv_data.email.clone()),
-			external_user_id: csv_data.email,
+			external_user_id: hex::encode(csv_data.email),
```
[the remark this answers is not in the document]
jannden: Yeah, I will squash the commits anyway, just kept them separate temporarily for you for better visibility.

tests/e2e.rs
jannden: Yeah, I suggest a separate PR. I am actually not sure how to log a user in with SSO here. Haven't seen such a flow as far as I remember. Maybe you can give it a try.

tlater-famedly (approved):

PR #100, opened 2024-12-13
feat: Add localpart to CSV import

### LINKED ISSUE #96: CSV: Add field to set userId & localpart
### Description
To migrate our remaining customers from LDAP to Zitadel, we need to reuse the localpart. 

### Solution
Add an extra field in the csv which is used to set the userId and localpart usermtadata.

jannden:
Closes #96 

Changes:
1. Added localpart field to the User struct as an optional field
2. Modified CSV import to store the localpart from CSV as metadata as localpart
3. Updated creation of new users in Zitadel to use localpart as Zitadel user ID
4. Emitting warning when attempting to update a user with a different localpart
5. Updated unit and e2e tests to verify


emgrav (approved):
Looks good, many thanks for getting this done so fast! Should we also tag a release in the same PR?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "famedly-sync"
-version = "0.8.0"
+version = "0.8.1"
```
emgrav, line 3:
```suggestion
version = "0.9.0"
```

emgrav (approved):

PR #106, opened 2025-01-20
refactor: Switch from ldap_poller to a direct implementation

tlater-famedly:
This cuts out an incredible amount of middle-man code that is completely unnecessary for famedly-sync in practice.

Also cleans up a couple of inherited bugs (like not correctly quitting if we run into a connection issue), and hopefully increases logging capacity. Basically, a prerequisite of famedly/ldap-poller#36, since this lets us reasonably fix the issue downstream.

TODO:

- [x] Rewrite the currently `sync` bits with async code
  - Apparently `ldap3` created a sync interface by just calling `block_on`-type things on their async methods, so it is completely incompatible with otherwise async code
- [ ] Check if this actually gives us more logs
  - Probably not actually necessary to merge this, it's good-to-have anyway
- [x] Figure out what the deleted entries thing is about again
- [x] Get the tests to pass
- [x] Write some new tests to assert the new bits work well
- [x] Cover error cases

### DIFF
--- a/src/sources/ldap.rs
+++ b/src/sources/ldap.rs
@@ -1,16 +1,13 @@
 //! LDAP source for syncing with Famedly's Zitadel.
 
-use std::{fmt::Display, path::PathBuf};
+use std::{fmt::Display, path::PathBuf, time::Duration};
 
 use anyhow::{anyhow, bail, Context, Result};
 use async_trait::async_trait;
-use ldap_poller::{
-	config::TLSConfig, ldap::EntryStatus, ldap3::SearchEntry, AttributeConfig, CacheMethod,
-	ConnectionConfig, Ldap, SearchEntryExt, Searches,
-};
+use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


src/sources/ldap.rs
```diff
@@ -199,56 +223,67 @@ pub struct LdapSourceConfig {
...
+				.server_certificate
+				.as_ref()
+				.map(std::fs::read)
+				.transpose()
+				.context("Failed to read server certificate")?
+				.map(|cert_data| Certificate::from_pem(cert_data.as_slice()))
+				.transpose()
+				.context("Invalid server certificate")?;
+
+			let identity: Option<Identity> = match (tls.client_key, tls.client_certificate) {
```
tlater-famedly, line 257: In retrospect, we should bundle these in a `ClientIdentity` attribute which is ~`Option<struct ClientIdentity(PathBuf, PathBuf)>` . That way we can assert at the type level that both are specified at the same time.

That'd require a breaking change though, so not doing it as part of this PR.

mzaniolo (approved):

PR #110, opened 2025-02-05
Use backend-build-workflows workflow for rust jobs

tlater-famedly:


tlater-famedly:
I figured out the issue: Since we clone *inside* the container, the repository is not actually available in the host's working directory. Should be pretty easy to fix with a copy in the nextest script.

tlater-famedly:
*Close* to working. To be clear, the current `sed` mess could be done better with environment variables, but it's currently hard to share env variables between setup scripts.

I mentioned this upstream, it seems they have implemented a fix since, it'd be nice to switch to that but I didn't want to break my local binary (nixpkgs doesn't package the new version yet): https://github.com/nextest-rs/nextest/issues/978#issuecomment-2649234741

PR #115, opened 2025-02-19
feat: Add filtering by org and project ids

### LINKED ISSUE #103: Syncs should be limited to the specified organization/project
Currently, when given a service user with sufficient permissions to do this, users who are not in the set organization/project will also be subjected to the sync, and therefore usually simply deleted. This is obviously very hazardous behavior, we should not just do that.

To limit the sync to just one organization is pretty simple:

```diff
diff --git a/src/zitadel.rs b/src/zitadel.rs
index ae205be..5c9c7e1 100644
--- a/src/zitadel.rs
+++ b/src/zitadel.rs
@@ -10,9 +10,10 @@ use zitadel_rust_client::{
 	v1::Zitadel as ZitadelClientV1,
 	v2::{
 		users::{
-			AddHumanUserRequest, IdpLink, InUserEmailsQuery, ListUsersRequest, Organization,
-			SearchQuery, SetHumanEmail, SetHumanPhone, SetHumanProfile, SetMetadataEntry,
-			TypeQuery, UpdateHumanUserRequest, User as ZitadelUser, UserFieldName, Userv2Type,
+			AddHumanUserRequest, AndQuery, IdpLink, InUserEmailsQuery, ListUsersRequest,
+			Organization, OrganizationIdQuery, SearchQuery, SetHumanEmail, SetHumanPhone,
+			SetHumanProfile, SetMetadataEntry, TypeQuery, UpdateHumanUserRequest,
+			User as ZitadelUser, UserFieldName, Userv2Type,
 		},
 		Zitadel as ZitadelClient,
 	},
@@ -95,9 +96,14 @@ impl Zitadel {
 	pub fn list_users(&mut self) -> Result<impl Stream<Item = Result<(User, String)>> + Send> {
 		self.zitadel_client
 			.list_users(
-				ListUsersRequest::new(vec![
-					SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human))
-				])
+				ListUsersRequest::new(vec![SearchQuery::new().with_and_query(
+					AndQuery::new().with_queries(vec![
+						SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human)),
+						SearchQuery::new().with_organization_id_query(OrganizationIdQuery::new(
+							self.zitadel_config.organization_id.clone(),
+						)),
+					]),
+				)])
 				.with_asc(true)
 				.with_sorting_column(UserFieldName::NickName),
 			)
```

However, limiting this to just the project grants is a bit more complex. This information is not tied to the user, but the project, so at the very least we would need an additional query for every user to check which projects they are in (i.e., search for the user's ID with https://zitadel.com/docs/apis/resources/mgmt/management-service-list-project-grant-members, *or* get a list of users with that search and add their IDs to the `AndQuery` - actually, the latter just seems better, hopefully there is not a limit to the number of IDs that can be added to a query).

These limits also add an issue in cases where an existing user is added to a new project, since it will create a new user instead of simply adding a new scope; this is especially problematic since we made the Zitadel ID a foreign key. I believe that that should never be the case in a famedly context, but we should confirm this.

sirewix:
Closes #103
Depends on [feat: add `list_project_grant_members` method by sirewix · famedly/zitadel-rust-client#55](https://github.com/famedly/zitadel-rust-client/pull/55)
Alternative to #114 as that approach seem to be impossible to implement in a reasonable time due to zitadel's [*Search User Grants*](https://zitadel.com/docs/apis/resources/mgmt/management-service-list-user-grants) not returning nickname (and phone)

### DIFF
--- a/.config/starte2e.sh
+++ b/.config/starte2e.sh
@@ -20,4 +20,5 @@ fi
 
 # Shut down any still running test-setup first
 docker compose --project-directory ./tests/environment down -v test-setup $ldap_down || true
-docker compose --project-directory ./tests/environment up --wait
+docker compose --project-directory ./tests/environment up --wait \
+	|| (docker compose --project-directory ./tests/environment logs test-setup; exit 1)
--- a/src/zitadel.rs
+++ b/src/zitadel.rs
@@ -1,5 +1,5 @@
 //! Helper functions for submitting data to Zitadel
-use std::path::PathBuf;
+use std::{path::PathBuf, pin::pin};
 
 use anyhow::{anyhow, bail, Context, Result};
 use base64::prelude::{Engine, BASE64_STANDARD};
@@ -15,10 +15,10 @@ use zitadel_rust_client::{
 		},
 		pagination::PaginationParams,
 		users::{
-			AddHumanUserRequest, AndQuery, IdpLink, InUserEmailsQuery, ListUsersRequest,
-			Organization, OrganizationIdQuery, SearchQuery, SetHumanEmail, SetHumanPhone,
-			SetHumanProfile, SetMetadataEntry, TypeQuery, UpdateHumanUserRequest,
-			User as ZitadelUser, UserFieldName, Userv2Type,
+			AddHumanUserRequest, AndQuery, IdpLink, InUserEmailsQuery, Organization,
+			OrganizationIdQuery, SearchQuery, SetHumanEmail, SetHumanPhone, SetHumanProfile,
+			SetMetadataEntry, TypeQuery, UpdateHumanUserRequest, User as ZitadelUser,
+			UserFieldName, Userv2Type,
 		},
 		Zitadel as ZitadelClient,
 	},
@@ -77,24 +77,23 @@ impl Zitadel {
 		&mut self,
 		emails: Vec<String>,
 	) -> Result<impl Stream<Item = Result<(ZitadelUserBuilder, String)>> + Send> {
-		self.zitadel_client
+		Ok(self
+			.zitadel_client
 			.list_users(
-				ListUsersRequest::new(vec![
+				Some(PaginationParams::DEFAULT.with_asc(true)),
+				Some(UserFieldName::NickName),
+				Some(vec![
 					SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human)),
 					SearchQuery::new().with_in_user_emails_query(
 						InUserEmailsQuery::new().with_user_emails(emails),
 					),
-				])
-				.with_asc(true)
-				.with_sorting_column(UserFieldName::NickName),
-			)
-			.map(|stream| {
-				stream.map(|user| {
-					let id = user.user_id().ok_or(anyhow!("Missing Zitadel user ID"))?.clone();
-					let user = search_result_to_user(user)?;
-					Ok((user, id))
-				})
-			})
+				]),
+			)?
+			.and_then(|user| async {
+				let id = user.user_id().ok_or(anyhow!("Missing Zitadel user ID"))?.clone();
+				let user = search_result_to_user(user)?;
+				Ok((user, id))
+			}))
 	}
 
 	/// Return a stream of Zitadel users
@@ -107,18 +106,16 @@ impl Zitadel {
 		Ok(self
 			.zitadel_client
 			.list_users(
-				ListUsersRequest::new(vec![SearchQuery::new().with_and_query(
-					AndQuery::new().with_queries(vec![
-						SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human)),
-						SearchQuery::new().with_organization_id_query(OrganizationIdQuery::new(
-							self.zitadel_config.organization_id.clone(),
-						)),
-					]),
-				)])
-				.with_asc(true)
-				.with_sorting_column(UserFieldName::NickName),
+				Some(PaginationParams::DEFAULT.with_asc(true)),
+				Some(UserFieldName::NickName),
+				Some(vec![SearchQuery::new().with_and_query(AndQuery::new().with_queries(vec![
+					SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human)),
+					SearchQuery::new().with_organization_id_query(OrganizationIdQuery::new(
+						self.zitadel_config.organization_id.clone(),
+					)),
+				]))]),
 			)?
-			.map(|user| {
+			.and_then(|user| async {
 				let id = user.user_id().context("Missing Zitadel user ID")?.clone();
 				let user = search_result_to_user(user)?;
 				Ok((user, id))
@@ -157,23 +154,18 @@ impl Zitadel {
 	/// Return a vector of a random sample of Zitadel users
 	/// We use this to determine the encoding of the external IDs
 	pub async fn get_users_sample(&mut self) -> Result<Vec<User>> {
-		let mut stream = self
+		let mut stream = pin!(self
 			.zitadel_client
 			.list_users(
-				ListUsersRequest::new(vec![
-					SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human))
-				])
-				.with_asc(true)
-				.with_sorting_column(UserFieldName::NickName)
-				.with_page_size(USER_SAMPLE_SIZE),
-			)
-			.map(|stream| {
-				stream.map(|user| {
-					let id = user.user_id().ok_or(anyhow!("Missing Zitadel user ID"))?.clone();
-					let user = search_result_to_user(user)?;
-					Ok((user, id))
-				})
-			})?;
+				Some(PaginationParams::DEFAULT.with_asc(true).with_page_size(USER_SAMPLE_SIZE)),
+				Some(UserFieldName::NickName),
+				Some(vec![SearchQuery::new().with_type_query(TypeQuery::new(Userv2Type::Human))])
+			)?
+			.and_then(|user| async {
+				let id = user.user_id().ok_or(anyhow!("Missing Zitadel user ID"))?.clone();
+				let user = search_result_to_user(user)?;
+				Ok((user, id))
+			}));
 
 		let mut users = Vec::new();
 
--- a/src/zitadel.rs
+++ b/src/zitadel.rs
@@ -3,16 +3,22 @@ use std::path::PathBuf;
 
 use anyhow::{anyhow, bail, Context, Result};
 use base64::prelude::{Engine, BASE64_STANDARD};
-use futures::{Stream, StreamExt};
+use futures::{Stream, StreamExt, TryStreamExt};
 use serde::{Deserialize, Serialize};
 use url::Url;
 use zitadel_rust_client::{
 	v1::Zitadel as ZitadelClientV1,
 	v2::{
+		management::{
+			V1UserGrantProjectIdQuery, V1UserGrantQuery, V1UserGrantRoleKeyQuery,
+			V1UserGrantUserIdQuery,
+		},
+		pagination::PaginationParams,
 		users::{
-			AddHumanUserRequest, IdpLink, InUserEmailsQuery, ListUsersRequest, Organization,
-			SearchQuery, SetHumanEmail, SetHumanPhone, SetHumanProfile, SetMetadataEntry,
-			TypeQuery, UpdateHumanUserRequest, User as ZitadelUser, UserFieldName, Userv2Type,
+			AddHumanUserRequest, AndQuery, IdpLink, InUserEmailsQuery, ListUsersRequest,
+			Organization, OrganizationIdQuery, SearchQuery, SetHumanEmail, SetHumanPhone,
+			SetHumanProfile, SetMetadataEntry, TypeQuery, UpdateHumanUserRequest,
+			User as ZitadelUser, UserFieldName, Userv2Type,
 		},
 		Zitadel as ZitadelClient,
 	},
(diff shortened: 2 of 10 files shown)


src/zitadel.rs
```diff
@@ -92,24 +97,66 @@ impl Zitadel {
...
+								},
+								V1UserGrantQuery::UserId {
+									user_id_query: V1UserGrantUserIdQuery::new()
+										.with_user_id(user_id),
+								},
+							]),
+						)?
+						.next()
+						.await;
+					Ok(grant.map(|_| x))
```
tlater-famedly, line 154:
```suggestion
					Ok(grant.is_some().then_some(x))
```

tests/environment/config.template.yaml
```diff
@@ -1,9 +1,10 @@
 zitadel:
   url: http://localhost:8080
   key_file: tests/environment/zitadel/service-user.json
-  organization_id: @ORGANIZATION_ID@
-  project_id: @PROJECT_ID@
-  idp_id: @IDP_ID@
+  organization_id: ORGANIZATION_ID
```
tlater-famedly, line 4: I'm not sure I see why we're removing the template markers here; they're to make explicit that these values should be changed if the file is used in practice.

tests/environment/test-setup.sh
```diff
@@ -6,6 +6,7 @@ set -eu
 # data and then to do some basic initialization.
```
tlater-famedly, line 6: Can we break these improvements into a separate commit? They're pretty extensive at this point.

tests/environment/test-setup.sh
```diff
@@ -6,6 +6,7 @@ set -eu
...
 #
 # This is intended for test suite setup, don't use this in production.
+log() { echo "$@" 1>&2; }
```
tlater-famedly, line 9: Is this actually what we want? Do the messages still correctly show up in e.g. GitHub CI when they go to stderr, and don't incur ordering issues?

tests/environment/test-setup.sh
```diff
@@ -100,23 +103,24 @@ idp_id="$(zitadel_request 'management/v1/idps/ldap' POST --json @- <<EOF | jq --
...
+log "Updating Zitadel IDs"
 org_id="$(zitadel_request 'management/v1/orgs/me' GET | jq --raw-output '.org.id')"
 
-sed "s/@ORGANIZATION_ID@/$org_id/" /config.template.yaml > /environment/config.yaml
-sed "s/@PROJECT_ID@/$project_id/" -i /environment/config.yaml
-sed "s/@IDP_ID@/$idp_id/" -i /environment/config.yaml
+sed "s/ORGANIZATION_ID/$org_id/;
+     s/PROJECT_ID/$project_id/;
+     s/IDP_ID/$idp_id/;
+    " /config.template.yaml > /environment/config.yaml
```
tlater-famedly, line 142: Nice spot, there should be a shellcheck lint for this :)

src/zitadel.rs
```diff
@@ -92,24 +97,66 @@ impl Zitadel {
 	}
 
 	/// Return a stream of Zitadel users
+	#[allow(clippy::type_complexity)]
 	pub fn list_users(
 		&mut self,
-	) -> Result<impl Stream<Item = Result<(ZitadelUserBuilder, String)>> + Send> {
-		self.zitadel_client
+	) -> Result<Pin<Box<dyn Stream<Item = Result<(ZitadelUserBuilder, String)>> + Send + Sync>>> {
```
tlater-famedly, line 103: We could probably create an alias for `(ZitadelUserBuilder, String)` to pacify clippy a bit here, and generally make things a bit more readable. Could even make a proper struct for it so that we can access `user.id` instead of `user.1`.

tests/environment/test-setup.sh
tlater-famedly: I wonder if something like this would be even cleaner:

```bash
sed -f - /config.template.yaml > /environment/config.yaml <<EOF
s/ORGANIZATION_ID/$org_id/;
s/PROJECT_ID/$project_id/;
s/IDP_ID/$idp_id/;
EOF
```

At the very least it'd be easier to touch with editors.

src/zitadel.rs
```diff
@@ -92,24 +97,66 @@ impl Zitadel {
...
+				let id = user.user_id().context("Missing Zitadel user ID")?.clone();
+				let user = search_result_to_user(user)?;
+				Ok((user, id))
+			});
+
+		Ok(if self.zitadel_config.filter_by_project_id {
+			let zitadel = self.zitadel_client.clone();
+			let org_id = self.zitadel_config.organization_id.clone();
+			let project_id = self.zitadel_config.project_id.clone();
+			Box::pin(stream.try_filter_map(move |x| {
```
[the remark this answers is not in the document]
sirewix: I agree that it needs a better name, but it's not user, it's `(user, user_id)`

tests/environment/test-setup.sh
sirewix: It's so that it can be used inside functions. I didn't commit it, but during debugging I put some logging into `zitadel_request` function. We can't log into stdout there as it would be appended or prepended to json response output.

The output is ordered as long as all logging is done to stderr (which is the norm for a lot of software, actually). We only log this output in case the test setup container fails

tests/environment/test-setup.sh
```diff
@@ -46,39 +48,40 @@ zitadel_request() {
...
 
-	curl -s \
+	curl -sS \
+		--fail-with-body \
 		--request "$_request_type" \
 		--url "${ZITADEL_HOST}/${_path}" \
 		--header 'Host: localhost' \
 		--header "Authorization: Bearer ${zitadel_token}" \
-		"$@"
+		"$@" || exit 1
```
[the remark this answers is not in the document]
sirewix: This would imply switching from `sh` to `bash`. Looks like `-o pipefail` and even `-e` are quite controversial

src/zitadel.rs
sirewix: This could have been avoided if zitadel returned user with strict id field. The alias you suggest I'd name something like `ZitadelUserWithStrictId`. Having a pair `(user, id)` is a workaround with minimum extra code added to make it work. 

I honestly don't know why clippy is complaining, to me the type is relatively simple, semantically it is even simpler: a result, returning a stream, which elements are this pair. If we remove necessary rust boilerplate, it would be `Result<Stream<Item = Result<(User, String)>>>`, which doesn't look to me complex and I'd rather turn off the lint. After all, lints are not laws, they are suggestions, pacifying every lint for the sake of it is harmful IMO

tests/environment/test-setup.sh
tlater-famedly: There are some logs you get from things like `ldapsearch` which will go to `stdout`, unfortunately. Conversely, there's also a bunch of software out there that will put its output on stderr to solve this problem.

There are no perfect solutions here, but yep, fair enough, this will probably save someone 10 minutes of headache sometime.

tests/environment/test-setup.sh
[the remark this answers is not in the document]
tlater-famedly: Oh my, I didn't even realize `pipefail` isn't POSIX, fair enough.

tests/environment/config.template.yaml
tlater-famedly: Ugh, can we rename the file to `.yaml.template` or explicitly exclude it from the pre-check or such?

.config/starte2e.sh
```diff
@@ -20,4 +20,5 @@ fi
 
 # Shut down any still running test-setup first
 docker compose --project-directory ./tests/environment down -v test-setup $ldap_down || true
-docker compose --project-directory ./tests/environment up --wait
+docker compose --project-directory ./tests/environment up --wait \
```
tlater-famedly, line 23: My point kind of was that we should probably keep all the test changes in one commit. This has little to do with filtering org and project ids.

src/zitadel.rs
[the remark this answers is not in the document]
tlater-famedly: Yep, fair, that would have been what I'd do too, but I figured I'd stay with the spirit of the existing code.

src/zitadel.rs
tlater-famedly: I can see where the lint is coming from, the boilerplate types are actually quite the mouthful, and the anonymous tuple makes it hard to spot at a glance what the function is supposed to do. If the lint is implemented well it'll have a heuristic that specifically penalizes us for using anonymous tuples.

`(ZitadelUserBuilder, String)` *is* a workaround and not trivial to grok if you're using the function without much context, particularly because `String` doesn't exactly scream "id". Adding another struct would make the intent of this function much clearer, and is only 3-4 extra lines, and no extra runtime overhead. We're not trying to code golf here, we probably shouldn't be using anonymous tuples unless there's a *really* good reason or the types are indeed trivial.

clippy might still complain either way because of all the boilerplate typing, but the lint did prompt me to consider whether an anonymous tuple really is the best way to represent that, and I've since decided that I don't think it is, so I'd say it's doing its job correctly.

But yep, it can be silenced, I just realized I have a preference for not doing so in this case because it's actually pointing at something that is a bit unclear. It's ultimately your call of course ;)

tests/environment/config.template.yaml
```diff
@@ -1,9 +1,10 @@
...
+  project_id: PROJECT_ID
+  idp_id: IDP_ID
+  filter_by_project_id: true
```
tlater-famedly, line 7: I think the only thing left to do is to write a test that sets this to false.

That said, I'm still not sure whether we actually need to make this configurable, or if we should just eat the performance penalty. This will confuse users, and I'm not looking forward to explaining it to 3 different people individually because they just copied the sample config (or saw an error message about a missing key).

We've conclusively determined that there's no better way to implement this, and in practice this is a safety feature to guard against misuse, so it should probably never be disabled to prevent accidents.

I think it'd be better to remove the option to turn this off, keep this as a feature branch to fulfill the immediate need, publish an off-version release with it, and then do some upstream work to make this possible to implement with reasonable performance characteristics. That, or we just merge it fully and try implementing the fancy new features for remote user integration for our next release.

What are your thoughts?

tlater-famedly (commented):
Almost there!

tests/environment/config.template.yaml
sirewix: Renaming to template would make editors not highlight the syntax, having it in pre-check has value. I think just quoting would be the best approach then

tests/environment/test-setup.sh
sirewix: My concern was to make curl error messages ordered with logs. I wonder how search results and error results live in one pipe, the way curl does it makes much more sense to me

src/zitadel.rs
sirewix: > the anonymous tuple makes it hard to spot at a glance what the function is supposed to do

I fully agree with you, but my objection here is that imo this type is not complex enought. Something like `Vec<Vec<(u8, String)>, HashMap<String, HasMap<String, String>>> is indeed unreadable, but in this case the `Item` type is marked with `Item = ...` So there's a clear boundary between "outer type" (result with boxed stream) and "inner type" (element of the stream) and that's why this particular lint doesn't make sense to me.

> If the lint is implemented well it'll have a heuristic that specifically penalizes us for using anonymous tuples

That sounds a lot like removing tuples from the language altogether, but they are part of the language, why would it make sense to penalize for using them?

> `(ZitadelUserBuilder, String)` _is_ a workaround and not trivial to grok if you're using the function without much context, particularly because `String` doesn't exactly scream "id".

I agree, but it's not me who implemented it this way, and I tried to implement this PR without too much extra changes unless I really can't resist them :D. First, I think order should be reversed, i.e. `(String, User)`, that would make it compatible with iterators that are made of or can be collected into maps (`HashMap`, `BTreeMap`). Second, yes, aliasing `type UserId = String` is great idea, I'd implement that. (A good practice could even be to make it even stricter with `struct UserId(String)`)

> Adding another struct would make the intent of this function much clearer, and is only 3-4 extra lines, and no extra runtime overhead.

I don't think we should add this struct because there are other simpler ways to make it clearer (like `(UserId, User)`) and because it is a workaround and I can't even think of proper name for it. `ZitadelUserBuilderWithUserIdExtracted`? Such long descriptive name would only add cognitive load on the code reader and lengthen the type signature.

> we probably shouldn't be using anonymous tuples unless there's a _really_ good reason or the types are indeed trivial.

Can't agree with you here, they are part of the language, they are a great tool for structural typing which has lots of benefits compared to nominal (tagged) typing. Should not be overused, ofcourse, but for little workaroundish value juggling I think it's ok.

The context here is that the type was (similar to `get_users_by_email`, btw)
```rust
	) -> Result<impl Stream<Item = Result<(ZitadelUserBuilder, String)>> + Send> {
```
all I did is boxed it 
```rust
	) -> Result<Pin<Box<dyn Stream<Item = Result<(ZitadelUserBuilder, String)>> + Send>>> {
```
So IMO added `Pin<Box<` around the *stream*, that is not that much of a complexity that should affect the other part of the signature (*stream element*)

.config/starte2e.sh
sirewix: Yep, forgot to move this change into that commit, fixed

tests/environment/config.template.yaml
sirewix: This is a good question. I think we can leave it configureable, but instead of defaulting to `false` as it's done now, we can default to `true` instead. This way in case of any unexpected bugs or performance issues for users that don't need this feature it could be turned off

tests/environment/config.template.yaml
sirewix: Looks like `<...>` doesn't trigger the linter, updated to that, seems like a good alternative to `@...@`

sirewix: Another option would be to use `envsubst` and `${...}` syntax

tlater-famedly: envsubst is overkill here, we don't actually need to protect these secrets. Current state should be fine.

tests/environment/config.template.yaml
tlater-famedly: I agree leaving it on by default is better, but I'm still not sure it's worth the maintenance overhead to have the option in the first place. Keeping the option technically necessitates a *lot* more testing, and the code to swap between two "modes" is itself not super clean. I'd prefer to just get rid of all of that.

This version does a server-side sort anyway so my concern with memory use has been alleviated, at which point the only reason we're doing this is in case we're introducing bugs, which seems a little overly defensive to me.

src/zitadel.rs
tlater-famedly: I think we'll have to agree to disagree on this one - I still think that having properly tagged names would be best here, especially since it would remove a destructure -> repack step where we only use one value. And yes, anonymous tuples have their uses, especially in iterator chains, but I believe they should be avoided in function signatures.

The naming issue is the thing closest to convincing me, but I'm sure we could find a better name and probably clean up some code further up the stack to make this less redundant and cleaner - though at that point it's admittedly no longer a small change in an unrelated PR.

I'd be well on board with a `struct ZitadelUserId(String)`, too. We should make some time to clean up some of these internals at some point.

tests/environment/config.template.yaml
tlater-famedly: Feel free to just resolve this when ready to merge, by the way.

src/zitadel.rs
tlater-famedly: Feel free to just resolve this when ready to merge by the way.

tests/environment/config.template.yaml
sirewix: I don't have strong opinion here, let's make it non configurable then. As a bonus that will resolve the bikeshedding regarding that lint

tlater-famedly (approved):

PR #116, opened 2025-02-24
refactor: Refactor `get_next_zitadel_user` into stream combinators + rust 2024

sirewix:


src/zitadel.rs
```diff
@@ -361,128 +354,67 @@ impl Zitadel {
...
+		let email = human_user
+			.email()
+			.and_then(|human_email| human_email.email())
+			.with_context(|| mk_err("email address"))?
+			.clone();
+
+		let phone = human_user.phone().and_then(|human_phone| human_phone.phone()).cloned();
+		let localpart = self
+			.zitadel_client
+			.get_user_metadata(&id, "localpart")
```
tlater-famedly, line 391: Nice, this is so much cleaner!

src/lib.rs
```diff
@@ -146,17 +111,19 @@ async fn sync_users(config: &Config, sync_users: &mut VecDeque<User>) -> Result<
 	// are not in the list
 	sync_users.retain(|user| user.enabled);
 
-	let mut zitadel = Zitadel::new(config).await?;
+	let zitadel = Zitadel::new(config).await?;
 	let mut stream = pin!(zitadel.list_users()?);
+	// TODO: refine where we want to skip errors and where not
```
tlater-famedly, line 116: We probably should not skip errors in the general case. We have the following use cases:

- Simple, every day sync run in a cron job
  - Skipping errors means that database problems are never found. While the sync is robust to failures to a degree (the algorithm reaches eventual consistency), if a more permanent error arises this *should* be noticed, so we should error with an unclean exit code.
- UKT user deletion
  - Same as above, jobs run in a cron job which will not actively be monitored. If a user is not deleted, we should raise an error so someone actually notices (and does whatever is necessary to fix the error)
- User migration
  - If a user fails to be migrated, this will cause them to lose their chat history on the next sync. The person performing the migration should be informed of the failure, so that they can roll back the database and do whatever they need to do to fix the error.

That said, I think there is nuance. We should probably handle network errors with retries, and we should avoid aborting on *certain* user format issues (e.g. missing localpart or external ID were deliberately skipped, as it means that we've run into a manually created user).

For this PR it's probably fine to continue ignoring errors, since that's closest to the current behavior, though it'd be nice to have an unclean exit code when we run into them.

Rethinking how we handle errors more granularly can be broken into the first ticket for a frontend refactor, which we're currently planning soon (tm).

tlater-famedly (commented):
Looks nice thanks for cleaning this up :)

src/lib.rs
```diff
@@ -99,40 +62,55 @@ pub async fn perform_sync(config: &Config) -> Result<()> {
...
-		zitadel.delete_user(&zitadel_user.1).await?;
-	}
+async fn delete_users_by_email(
+	zitadel: &Zitadel<'_>,
+	skipped_errors: &SkippedErrors,
+	emails: Vec<String>,
+) -> Result<()> {
+	zitadel
+		.get_users_by_email(emails)?
+		.try_for_each_concurrent(Some(4), async |(zitadel_id, _)| {
```
tlater-famedly, line 82: Ah, so beautiful.

src/lib.rs
```diff
@@ -99,40 +62,55 @@ pub async fn perform_sync(config: &Config) -> Result<()> {
...
+			zitadel
+				.delete_user(&zitadel_id)
+				.await
+				.skip_zitadel_error("deleting user", skipped_errors);
```
tlater-famedly, line 86: Mostly thinking out loud here, but would it not make more sense to `.collect::<Vec<Result<_, Err<_>>>>`, and then return the `Vec<Err<_>>` instead of doing this? If we use a custom error enum we can add a `.is_skippable()` to identify whether to abort or not in the future, and just generally have nicer integration.

I assume you've attempted that route, it'd be nice to leave a comment somewhere to explain why this isn't possible.

sirewix: What would this achieve? I didn't thought on such improvement here, all we do here is delete, and delete operation seem like something we can safely skip (but return non zero exit code later)

sirewix: Commented out this line (and similar ones) as it breaks certain test cases and is breaking change.

PR #118, opened 2025-02-25
release: v0.10.0-rc1

sirewix:


emgrav (approved):

emgrav:
if we specifically release the release candidate like this rather than just tag the RC on the v0.10.0 release branch, we need to make sure to follow up with a proper release

PR #127, opened 2025-03-19
feat: Allow LDAP bin attributes to be parsed as utf8

sirewix:
Attributes that were configured to be binary, but we expect them to be strings would fail in runtime with "attribute X expected to be binary", which doesn't make much sense. This PR tries to parse binary attibutes as utf8.

Alternative to #126


tlater-famedly:
I'm not sure I follow *why* we want this.

If you want to process an attribute that LDAP can only report as binary, you should configure it as binary. This should *always* treat the value as binary, encoding it with hex encoding. If such a value can be decoded as UTF-8, that does not mean it's a UTF-8 value; it simply means that the binary value happens to be encodeable as UTF-8.

sirewix:
After discussion with @tlater-famedly, closing this in favor of #126. This feature may be desireable in the future but at the moment it needs a bit more refinement

PR #135, opened 2025-04-23
feat(install-ids): Add ID installation script

### LINKED ISSUE #130: Create an LDAP ID matching script
See famedly/infra-meta#2112.

Effectively, sometimes things go wrong in production, and we end up with deployments where users exist that do not have LDAP IDs matching the ones in a customer's LDAP deployment. Presumably this happens because of syncs to existing instances.

While our documentation does warn against doing that, obviously it still happens, so we clearly need some kind of recovery plan. A script that uses email addresses to match users to their LDAP IDs for a one-time equalization is probably useful (famedly-sync cannot do this in general since email addresses may change; LDAP IDs do not).

tlater-famedly:
Fixes #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,86 @@ Config can have **various sources** to sync from. When a source is configured, t
 
 ## Migrations
 
+### Existing Zitadel deployments
+
+For Zitadel deployments that have not been synced to before, we
+provide an "ID installation" binary. This links users to their LDAP
+counterparts via the LDAP ID, so that famedly-sync can correctly
+identify them in subsequent sync runs.
+
+This is intended to run *exactly* once, when initializing an existing
+Zitadel instance for being synced to with famedly-sync in the future.
+
+Ideally, this should never need to be done, however in practice we
+have quite a few legacy instances, which could not be migrated
+without adequate tooling.
+
+The tool expects a 1:1 mapping from Zitadel email addresses to LDAP
+users. Zitadel users must be in the configured organization and
+project, as well as have the `User` grant.
+
+There are a couple of known failure conditions:
+
+- If any admin users were created, and additionally given the `User`
+  grant, the service user will likely not be able to update them
+- Users without email addresses will be skipped
+- If multiple users share an email address, they will not be silently
+  given the same IDs
+- Users without corresponding LDAP users will be skipped
+
+Any of these errors, as well as more unpredictable scenarios, will be
+logged with the associated user's Zitadel ID. Should such errors
+occur, a manual migration of these users is likely necessary - for
+this, simply use the Zitadel UI to change the user's `Nickname` field
+to the hex-encoded value of their LDAP id. To convert the string
+reported by `ldapsearch`, this python script should be enough:
+
+```python
+import base64
+import sys
+
+ldapsearch_id = sys.argv[1]
+
+uid = base64.standard_b64decode(ldapsearch_id).hex()
+print(uid)
+```
+
+The tool performs the following actions for each Zitadel user:
+
+ 1. Gather the current list of LDAP users according to the famedly-sync
+    filter configuration
+ 2. Start iterating through all Zitadel users, taking their email address
+    and nickname
+    1. If no nickname is set, check the list of LDAP users for users with
+       matching email addresses
+       - If exactly one user with this address exists, take their ID
+         attribute, encode it appropriately, and write it to the Zitadel
+         user's nickname
+       - Otherwise, print a warning and continue with the next user
+    2. If a nickname is already set, double check that email address is
+       unique among Zitadel users and that the nickname matches LDAP's ID
+       attribute
+      - If not, print a warning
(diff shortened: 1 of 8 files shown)


src/bin/install-ids.rs
```diff
@@ -0,0 +1,34 @@
+//! Script to install LDAP IDs into Zitadel users whose email
```
tlater-famedly, line 1: My attempt at docs-driven design in lieu of a clearer place to put this for now.

src/zitadel.rs
```diff
@@ -106,11 +106,11 @@ impl<'s> Zitadel<'s> {
 			.filter_map(Skippable::filter_out))
 	}
 
-	/// Return a stream of Zitadel users
+	/// Return a stream of raw Zitadel users
 	#[tracing::instrument(skip_all)]
-	pub fn list_users(
+	pub fn list_users_raw(
```
tlater-famedly, line 111: We should probably refactor this whole module and make all other functions call into this one, and give it a *few* more args. famedly-sync should only ever operate on users with the set group.

That said, we're on the cusp of no longer directly talking to Zitadel anyway, so maybe trying a big refactor here is pointless.

tests/common/mod.rs
```diff
@@ -0,0 +1,216 @@
+//! Common test helpers
```
tlater-famedly, line 1: The `mod.rs` name in a subdirectory is mandatory for tests, since putting a module entrypoint in the top-level test directory results in it being treated as a test file.

src/bin/install-ids.rs
sirewix: This should be in the readme as well. My suggestion would be to either
1. mk a separate install-ids-readme.md and `#![doc = include_str!(install-ids-readme.md)]
2. keep it here, but use `mdsh` to `grep '\/\/!' src/bin/install-ids.rs | sed <strip //!` to include it into the README.md

src/lib.rs
```diff
@@ -253,6 +257,112 @@ async fn sync_users(
...
+		let users = ldap_client.get_sorted_users().await.context("Failed to query LDAP users")?;
+		users.into_iter().map(|user| (user.email.clone(), user)).collect()
+	};
+
+	let mut zitadel_users =
+		pin!(zitadel_client.list_users_raw().context("failed to query Zitadel users")?);
+
+	let mut seen_emails: HashSet<String> = HashSet::new();
+
+	while let Some(user) = zitadel_users.next().await.transpose().context("failed to query user")? {
```
sirewix, line 280: To avoid tracing::error + continue, my suggestion would be to move the whole while loop body into a separate function and just return the errors there (with `.context` or whaterver). Then here we can just 
```rust
while let Some(user) = ... {
  let _ = link_zitadel_user(user).await.inspect_err(|e| tracing::error!("{e:?}"));
}
```

sirewix: Another idea is to use [`futures::stream::StreamExt::for_each_concurrent`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each_concurrent) instead of the loop. The only caveat is that you'd need locking for `seen_emails` but it looks like it's not really necessary anyway

src/bin/install-ids.rs
sirewix: And the README.md should contain a section on how (and when/why) to run this, something like

> If there's ... this ... issue ..., there's `install-ids` binary, to run it run `cargo run --bin install-ids`, to run it with docker `docker run -v config.yaml:config.yaml --rm famedly-sync install-ids`

src/lib.rs
tlater-famedly: I don't think we want to avoid `tracing::error` + continue. This script will be used against instances on which we have existing users that have been created in haphazard ways. Errors will likely almost always occur. We still want to set IDs for the remaining users - since there is no ordering concern here, this isn't an issue at all.

We *could* reuse the `SkippableError` thing instead, but I'd rather not - there's no reason to make this more complex with that.

README.md
```diff
@@ -43,6 +43,53 @@ Config can have **various sources** to sync from. When a source is configured, t
 
 ## Migrations
 
+### Existing Zitadel deployments
+
+For Zitadel deployments that have not been synced to before, we
+provide an "ID installation" binary. This links users to their LDAP
+counterparts via the LDAP ID, so that famedly-sync can correctly
+identify them in subsequent sync runs.
```
emgrav, line 51: I would still include up here right at the beginning that this migration is intended to be used EXACTLY ONCE, as well as the criteria for determining whether this should be used.

emgrav: Also some information ab out what can go wrong when using it and how to respond to problems

emgrav (approved):

src/lib.rs
sirewix: Here's a PR: https://github.com/famedly/famedly-sync/pull/137

PR #139, opened 2025-04-25
feat(contrib): Add a helper script for ldapsearch debugging

tlater-famedly:
We've needed to figure out what famedly-sync is seeing a few times while debugging in customer environments now; this is a quick solution that helps get this more easily.

Short of a full command to do various searches in famedly-sync (which would basically just reimplement `ldapsearch`) I think this is a pretty good way to patch up that use case a bit.

### DIFF
--- a/contrib/extract-ldapsearch-command.sh
+++ b/contrib/extract-ldapsearch-command.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+#
+# A small script to convert famedly-sync configuration into its
+# equivalent `ldapsearch` command; this can help find out what
+# `famedly-sync` sees in practice when performing a sync.
+#
+# Usage:
+#
+# ./extract-ldapsearch-command.sh CONFIG
+
+config="$1"
+
+if ! [ -e "$config" ]; then
+	echo "Please give the famedly-sync configuration file as the first argument"
+	exit 1
+fi
+
+grep_from_yaml() {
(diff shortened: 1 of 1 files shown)


jannden (approved):
Looks fine

contrib/extract-ldapsearch-command.sh
sirewix, line 1: what is `./contrib/`?

tlater-famedly: Just a directory with some random bonus scripts that are useful. It's pretty common in FOSS projects; often plugins and other related things not part of the main codebase make it in there, too.

contrib/extract-ldapsearch-command.sh
```diff
@@ -0,0 +1,36 @@
...
+
+config="$1"
+
+if ! [ -e "$config" ]; then
+	echo "Please give the famedly-sync configuration file as the first argument"
+	exit 1
+fi
+
+grep_from_yaml() {
+	grep "$1" "$config" | cut -f 2 -d ':' | tr -d '" '
```
tlater-famedly, line 19: Sadly the simple approach doesn't work very well in practice; various attributes used in DNs and filters often contain spaces and `:`. `tr`-ing and `cut`-ing *all* of them breaks.

We'd need to do a more thorough yaml parse to not run into this, sadly non-greedy matches are hard to do with just POSIX utilities.

Issue #143, opened 2025-05-15
Don't delete users when they are deleted in LDAP


tlater-famedly:
We've frequently [run into issues](https://github.com/famedly/product-management/issues/3124#issuecomment-2880630928) with mass user deletions because of small changes in filters and straight-up bugs or mistakes. Since propagating a deletion is very destructive, this is generally quite bad.

It appears that in practice hospitals do not actually delete users, but only set the disable flag; to prevent these issues for the time being, we should only delete users when the disable flag is set on them, and pretend that any other missing users are just mistakes.

A longer-term solution is #142.

PR #149, opened 2025-07-04
feat: Delete only disabled users

jannden:
Closes https://github.com/famedly/famedly-sync/issues/143

### DIFF
--- a/src/zitadel.rs
+++ b/src/zitadel.rs
@@ -472,7 +472,7 @@ struct Skippable;
 
 impl Skippable {
 	/// Helper function to use with `[StreamExt::filter_map]`
-	#[allow(clippy::unused_async, clippy::collapsible_if)]
+	#[allow(clippy::unused_async)]
 	#[tracing::instrument(skip_all)]
 	pub async fn filter_out<T: Send>(res: Result<T>) -> Option<Result<T>> {
 		// should we `skipped_errors.notify_error()` here?


src/lib.rs
```diff
@@ -151,72 +147,89 @@ async fn sync_users(
...
+						new_user.external_user_id
+					);
+				} else {
+					zitadel
+						.import_user(&new_user)
+						.await
+						.with_context(|| {
+							format!("Failed to import user `{}`", new_user.external_user_id)
+						})
+						.skip_zitadel_error("importing user", skipped_errors);
```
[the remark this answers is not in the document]
jannden: Yes, see [here](https://github.com/famedly/famedly-sync/blob/main/src/zitadel.rs#L490) and [here](https://github.com/famedly/famedly-sync/blob/main/src/lib.rs#L374).

src/lib.rs
```diff
@@ -151,72 +147,89 @@ async fn sync_users(
...
-				zitadel
-					.import_user(&new_user)
-					.await
-					.with_context(|| {
-						format!("Failed to import user `{}`", new_user.external_user_id)
-					})
-					.skip_zitadel_error("importing user", skipped_errors);
+				if !new_user.enabled {
+					tracing::warn!(
+						"Excess sync source user `{}` is disabled, skipping import",
```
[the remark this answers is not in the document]
jannden: I added the clearer version of the message and changed to `debug`

src/lib.rs
```diff
@@ -151,72 +147,89 @@ async fn sync_users(
...
 
 			// If the sync source user matches the Zitadel user, the
-			// user is already synced and we can move on
-			(Some(new_user), Some((_, existing_user))) if new_user == existing_user => {
+			// user is already synced
+			(Some(new_user), Some((zitadel_id, existing_user))) if new_user == existing_user => {
+				if !new_user.enabled {
+					// If the user is disabled in the source, delete from Zitadel
+					// This will retroactively remove disabled users
+					zitadel
```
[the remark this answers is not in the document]
jannden: Added

src/lib.rs
```diff
@@ -228,14 +241,24 @@ async fn sync_users(
...
-				zitadel
-					.update_user(&zitadel_id, &existing_user, &new_user)
-					.await
-					.with_context(|| {
-						format!("Failed to update user `{}`", new_user.external_user_id,)
-					})
-					.skip_zitadel_error("updating user", skipped_errors);
-
+				if !new_user.enabled {
+					// If the user is disabled in the source, delete from Zitadel
```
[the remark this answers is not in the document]
jannden: Added

src/zitadel.rs
```diff
@@ -258,6 +258,42 @@ impl<'s> Zitadel<'s> {
...
+					);
+
+					// Look up the existing user by email
+					let mut existing_users =
+						pin!(self.get_users_by_email(vec![imported_user.email.clone()])?);
+
+					if let Some((existing_zitadel_id, existing_user)) =
+						existing_users.next().await.transpose()?
+					{
+						tracing::debug!(
```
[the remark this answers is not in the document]
jannden: Changed to `info`

src/zitadel.rs
```diff
@@ -258,6 +258,42 @@ impl<'s> Zitadel<'s> {
...
+							existing_user.external_user_id,
+							imported_user.external_user_id
+						);
+
+						// Update the existing user with the new external ID and other changes
+						self.update_user(&existing_zitadel_id, &existing_user, imported_user)
+							.await?;
+					} else {
+						// This shouldn't happen, but if it does, re-throw the original error
+						tracing::debug!(
```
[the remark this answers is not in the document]
jannden: Good call

src/zitadel.rs
```diff
@@ -258,6 +258,41 @@ impl<'s> Zitadel<'s> {
 				if error.to_string().contains("PHONE-so0wa") {
 					user.reset_phone();
 					self.zitadel_client.create_human_user(user).await?;
+				} else if error.to_string().contains("User already exists") {
```
sirewix, line 261: Is there a error code for this? (Similar to `PHONE-so0wa` above)

src/zitadel.rs
```diff
@@ -258,6 +258,41 @@ impl<'s> Zitadel<'s> {
...
+					// Handle the case where a user with the same email already exists
+					// This can happen when the external ID changes but the email stays the same
+					// Since we are keeping deleted users in Zitadel for safety reasons unless they
+					// are explicitly disabled in LDAP, we need to update the Zitadel user instead
+					// of how we did it previously (deleting old and creating a new one)
```
sirewix, line 266: *Are* we keeping deleted users in zitadel? I see `zitadel.delete_user(...)` calls in the sync code.

src/zitadel.rs
```diff
@@ -258,6 +258,41 @@ impl<'s> Zitadel<'s> {
...
+
+						// Update the existing user with the new external ID and other changes
+						self.update_user(&existing_zitadel_id, &existing_user, imported_user)
+							.await?;
+					} else {
+						// This shouldn't happen, but if it does, re-throw the original error
+						anyhow::bail!(error.context(format!(
+							"Failed to find existing user having different external ID {} by email",
+							imported_user.external_user_id
+						)));
```
sirewix, line 294: Not sure if it makes sense to use upstream error here, might look confusing in the logs. Why not just `existing_users.next().context("failed to find existing user ...")?`?

src/zitadel.rs
jannden: The idea is to only delete users when the disable flag is set on them, and pretend that any other missing users are just mistakes.

src/zitadel.rs
jannden: Makes sense

src/zitadel.rs
```diff
@@ -263,6 +263,40 @@ impl<'s> Zitadel<'s> {
...
+
+					tracing::info!(
+						"Found existing user with Zitadel ID {} by email, updating from external ID {} to {} and possibly other changes",
+						existing_zitadel_id,
+						existing_user.external_user_id,
+						imported_user.external_user_id
+					);
+
+					// Update the existing user with the new external ID and other changes
+					self.update_user(&existing_zitadel_id, &existing_user, imported_user).await?;
```
sirewix, line 299: Might make sense to put `.with_context(...)` and `.skip_zitadel_errror()` to some of these new pieces of code. But I think the skips were introduced initially to preserve previously existing behavior, I'm not sure if there are policy/requirements on what to do in case of errors

sirewix (dismissed):

src/zitadel.rs
sirewix: Yeah, but what about `skip_zitadel_error()`s?

src/zitadel.rs
```diff
@@ -438,7 +472,7 @@ struct Skippable;
 
 impl Skippable {
 	/// Helper function to use with `[StreamExt::filter_map]`
-	#[allow(clippy::unused_async)]
+	#[allow(clippy::unused_async, clippy::collapsible_if)]
```
tlater-famedly, line 475: I really don't think we should allow collapsible ifs, where do we use one?

tlater-famedly (dismissed):

src/zitadel.rs
jannden: Yeah nice, will remove that. It was due to [this run](https://github.com/famedly/famedly-sync/actions/runs/16068839283/job/45348690609), which since started working with collapsible if refactored in [another commit](https://github.com/famedly/famedly-sync/commit/28ec64de36fcaade70086f6cb80ab7837f558b4d#diff-3e25b405a8d29861116c7080aec6a67576e904d1885a1c871997b2a479933259R440-R444) on top of which this was later rebased.

src/lib.rs
```diff
@@ -151,72 +147,94 @@ async fn sync_users(
...
 				if new_user.external_user_id < existing_user.external_user_id =>
 			{
-				zitadel
-					.import_user(&new_user)
-					.await
-					.with_context(|| {
-						format!("Failed to import user `{}`", new_user.external_user_id,)
-					})
-					.skip_zitadel_error("importing user", skipped_errors);
+				if !new_user.enabled {
```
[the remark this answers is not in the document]
tlater-famedly: Yeah, I meant actually doing the bigger change. I think it'd help with readability - reducing the levels of indentation is almost always a net benefit. In this case especially; Whether or not a user is disabled has effectively become an essential part of the "what should we do with this user" logic.

PR #153, opened 2025-08-14
feat: Adjust for new UKT OAuth2 response

jannden:
UKT changed their OAuth2 response, now it doesn't include `id_token`


src/sources/ukt.rs
```diff
@@ -307,23 +299,47 @@ mod tests {
 	#[tokio::test]
 	#[ignore]
 	/// Connects to the real URL in config to get the OAuth2 token
+	/// You need to have a `.env` file with UKT_CLIENT_ID and UKT_CLIENT_SECRET
```
tlater-famedly, line 302: Hmm, I'm never a fan of the dotenv shenanigans.

I understand the point here is to have a test which runs against their real endpoint, but I'm also unsure if this is a good idea to begin with. Tests that depend on external infrastructure are notoriously flaky and difficult to run. We won't run them in practice.

I'd much prefer if we simply mocked their API, (mostly) statically recording the responses as they happen in practice today; this would also document the extant requirements, as well as give us proof should their API change (again).

src/sources/ukt.rs
```diff
@@ -307,23 +299,47 @@ mod tests {
 	#[tokio::test]
 	#[ignore]
```
tlater-famedly, line 300: This is an aside, since this PR doesn't introduce this (and we didn't have the infrastructure in place to do things right when these tests were originally written), but if this continues to depend on the external API we should exempt it using nextest rather than abusing `#[ignore]`.

tlater-famedly (changes requested):

tlater-famedly (approved):

PR #158, opened 2025-12-16
feat(zitadel): Remove the feature setting custom Zitadel IDs

tlater-famedly:
This feature is purely cosmetic, we've avoided using it for anything; As expected, a consistency bug was caused by it (likely due to bad handling of the historic status of this as an arbitrarily generated ID), though the bug happened upstream in Zitadel.


mzaniolo (dismissed):
Looks good to me

emgrav (approved):

PR #163, opened 2026-05-29
[FC-523] fix(ldap): page search results to avoid sizeLimitExceeded

### LINKED ISSUE #162: Sync aborts with LDAP `sizeLimitExceeded` (rc=4) when result set crosses server cap
## Summary

`famedly-sync` aborts the entire sync with `LDAP operation result: rc=4 (sizeLimitExceeded)` once the LDAP result set grows past the directory server's per-request entry cap. After the failure no users are reconciled with Zitadel until somebody intervenes; the sync is fully blocked, not partially degraded.

A representative log excerpt:

```
ERROR famedly_sync: perform_sync at /app/src/lib.rs:63:31
Caused by:
    0: Failed to query users from LDAP at `/app/src/lib.rs@36:5`
    1: LdapSource::get_sorted_users at src/sources/ldap.rs:49:40
    2: LDAP operation result: rc=4 (sizeLimitExceeded), dn: "", text: ""
```

## Root cause

`LdapSource::get_sorted_users` issues a single non-paged search and treats anything other than `success` as a hard error:

```rust
let (search_results, _stats) = ldap
    .search(
        &self.ldap_config.base_dn,
        Scope::Subtree,
        &self.ldap_config.user_filter,
        self.ldap_config.clone().get_attribute_list(),
    )
    .await?
    .non_error()?;
```

`rc=4` ([RFC 4511 §4.5.3](https://datatracker.ietf.org/doc/html/rfc4511#section-4.5.3)) is exclusively server-enforced and is returned when an unpaged search would exceed the server's per-request entry cap. Common vendor defaults that this falls foul of:

- **Active Directory** — `MaxPageSize` defaults to **1000**.
- **OpenLDAP `slapd`** — `sizelimit` defaults to **500**.
- **389 Directory Server / FreeIPA / RHDS** — `nsslapd-sizelimit` defaults to **2000**.
- **OpenDJ / ForgeRock DS** — `ds-rlim-size-limit` defaults to **1000**.

Any directory running with default settings will start failing the sync the moment the user cohort crosses the relevant threshold. The failure is sharp and binary — there is no partial sync, no degraded mode, and partial result entries the server may have sent before the rc=4 are discarded by `.non_error()?`.

## Impact

- AD-backed deployments hit this whenever the synced cohort grows past `MaxPageSize` (default 1000).
- OpenLDAP-backed deployments are exposed at a much lower threshold (500 entries by default).
- Different DCs / replicas can have different per-policy caps, so a load-balanced setup can suddenly start failing if traffic is steered to a stricter server.
- Workarounds available to operators today are all unsatisfying: raise the server-side cap, narrow `user_filter` (which silently drops users), or split the directory.

## Proposed fix

Switch `LdapSource::get_sorted_users` to use the **LDAP Simple Paged Results control (RFC 2696, OID `1.2.840.113556.1.4.319`)** so result sets larger than the server's per-request cap are retrieved across multiple pages instead of aborting.

The `ldap3` crate already ships an `adapters::PagedResults` adapter; it transparently issues follow-up searches with the returned cookie until the server signals end-of-results. With paging enabled, AD's `MaxPageSize` becomes a **per-page** cap (which is what it's actually meant to be), not a hard ceiling on the cohort size, and OpenLDAP's `sizelimit` is bypassed for paged searches under typical configurations.

Page size is exposed as a `sources.ldap.page_size` config option that defaults to **500** — comfortably below the most common server-side caps — and can be set to `0` to opt out of paging entirely for the rare case where a directory misbehaves with the control.

## Linked PR

Implementation in #163.

## Verification

- Unit tests covering the new config field's default, an explicit override, and the disabled-paging value.
- The existing E2E suite already exercises `LdapSource` against the OpenLDAP test container and continues to pass; the new code path is therefore covered for every E2E scenario.
- Smoke-tested locally with `cargo check --all-targets`, `cargo clippy --all-targets`, `cargo fmt --check`, and `cargo test --lib sources::ldap::tests`.

## Out of scope

- AD `MaxQueryDuration` (default 120 s) is a sibling cap that bites long-running searches and is unrelated to entry counts; not addressed here.
- Future enhancement: surface paged-search progress in tracing logs (page count, cumulative entries) to make debugging large directories easier.

chrismafam:
Fixes #162.
Closes [FC-523]

## Summary

- `LdapSource::get_sorted_users` now performs a paged streaming search using the LDAP Simple Paged Results control (RFC 2696, OID `1.2.840.113556.1.4.319`) via the `ldap3` `adapters::PagedResults` adapter, instead of a single non-paged `search()`.
- A new `sources.ldap.page_size` config option controls the per-page entry count; it defaults to **500**, which sits comfortably below the most common server-side caps (Active Directory `MaxPageSize` default 1000, OpenLDAP `sizelimit` default 500). Setting it to `0` disables paging for the rare case where a directory misbehaves with the control.

## Why

Without paging, any LDAP server that returns more entries than its per-request cap responds with `rc=4 (sizeLimitExceeded)`. famedly-sync's `.non_error()?` turns that into a hard error, discarding any partial entries the server may have sent and aborting the entire sync. Issue #162 has the full diagnosis.

With Simple Paged Results, AD's `MaxPageSize` becomes a per-page bound (which is what it's actually meant to be), not a hard ceiling on the cohort size, and OpenLDAP's `sizelimit` is bypassed for paged searches under typical configurations. The `ldap3` adapter handles the cookie loop and follow-up searches transparently.

## Test plan

- [x] `cargo check --all-targets`
- [x] `cargo clippy --workspace --all-targets` — no new warnings; pre-existing `src/config.rs:101 FeatureFlags missing docs` and `Cargo.toml string_to_string` are unchanged from `main`.
- [x] `cargo fmt -- --check`
- [x] `cargo test --lib sources::ldap::tests` — all 7 tests pass, including 3 new ones:
  - `test_page_size_default` — verifies the field defaults to 500 when omitted from YAML.
  - `test_page_size_explicit_override` — verifies an explicit value (250) is honored.
  - `test_page_size_disabled` — verifies `page_size: 0` deserializes correctly.
- [x] Existing E2E suite (runs `LdapSource` against the OpenLDAP test container in `tests/environment/`) covers the new code path for every scenario it exercises.
- [ ] Reviewer to verify on a real AD directory if convenient — see "Manual smoke test" below.

## Manual smoke test (optional)

To confirm against a real LDAP server (any vendor):

1. Point an `ldap-config.yaml` at the directory and ensure the matching cohort is larger than the server's per-request cap (e.g. >1000 for AD with default `MaxPageSize`).
2. Run `cargo run --bin famedly-sync` against it.
3. Observe that `LdapSource::get_sorted_users` returns the full cohort and the sync completes, where it would previously have aborted with `rc=4`.

## Notes for review

- The change keeps the existing bind / unbind / connection-error-handling flow; only the search call site changes.
- The default of 500 was chosen so the same client config works against both AD (cap 1000) and an unconfigured OpenLDAP (cap 500) without further tuning. Increasing it on AD-only deployments (e.g. to 900) is safe and slightly reduces round-trips but isn't required.
- `EntriesOnly` is included in the adapter chain alongside `PagedResults` to keep referrals out of the entry stream and into `LdapResult.refs`, mirroring the prior behavior of `Ldap::search()`.
- Sample configs (`sample-configs/ldap-config.sample.yaml`, `sample-configs/ad-config.sample.yaml`) document the new option as a commented-out line so existing customers can ignore it but discover it when reading the samples.

Made with [Cursor](https://cursor.com)

[FC-523]: https://famedly.atlassian.net/browse/FC-523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/sources/ldap.rs
```diff
@@ -40,27 +42,35 @@ impl Source for LdapSource {
...
-		let mut users: Vec<User> = search_results
-			.into_iter()
-			.map(SearchEntry::construct)
-			.map(|entry| self.parse_user(entry))
-			.try_collect()?;
+		let mut users: Vec<User> = Vec::new();
+		while let Some(entry) = search.next().await? {
+			let entry = SearchEntry::construct(entry);
+			users.push(self.parse_user(entry)?);
+		}
```
sirewix, line 71:
```suggestion
		let mut users: Vec<User> = search
			.and_then(|entry| future::ok(self.parse_user(SearchEntry::construct(entry))))
			.try_collect()
			.await?;
```

mstoldt-famedly:
This ticket is covered in Jira https://famedly.atlassian.net/browse/FC-523

src/sources/ldap.rs
sirewix: nevermind, it doesn't implement `Stream`, it's some custom thing in `ldap3`

sample-configs/ad-config.sample.yaml
```diff
@@ -39,6 +39,13 @@ sources:
     user_filter: "(memberOf=cn=App)"
     # The LDAP connection timeout
     timeout: 5
+    # Page size for the LDAP Simple Paged Results control (RFC 2696).
```
mzaniolo, line 42: Could we also add this information to the readme?

sirewix: I'm not sure where I could put it, our configuration documenation is just a reference to config samples, commenting just one one parameter doesn't feel right. It should also work out of the box, the default value is 500 and it should be correct for the most AD implementations.

Speaking of documentation, we should probably add a generated json schema for our config

sirewix: I created a ticket: [FC-525](https://famedly.atlassian.net/browse/FC-525)

mzaniolo (dismissed):

Revanee (approved):
Looks good to me!

PR #164, opened 2026-06-20
release: v0.12.0 — migrate to famedly-zitadel-rust-client v0.12

nikzen:
## Summary
- Bump the Zitadel client from legacy `zitadel-rust-client` v0.3 to `famedly-zitadel-rust-client` v0.12 and migrate the e2e tests off the gRPC v1 API onto the v2 HTTP API, dropping the `v1` dev-dependency feature that carried the remaining security advisories (`cargo audit` now clean).
- Fix two error-handling regressions in `src/zitadel.rs` surfaced by running e2e against the v2 production code: the v2 client exposes Zitadel error codes (`V3-DKcYh`, `PHONE-so0wa`, `COMMAND-ieJ2e`) as a *source* in the error chain rather than in the top-level `Display`. Match against the full chain via `{error:#}` to restore the "user already exists" update path and the invalid-phone fallback in both `import_user` and `update_user`.
- Add a `ZitadelExt` test helper trait and rewrite the e2e suite to use v2 idioms (`Option` semantics, `human()`/`profile()`/`is_verified()` accessors, login-name lookup, grants, IDP links).
- Release **v0.12.0**: bump `Cargo.toml`/`Cargo.lock` and add a `CHANGELOG.md` entry.

Closes: https://famedly.atlassian.net/browse/FC-554


Cargo.toml
```diff
@@ -33,7 +33,7 @@ tracing-subscriber = "0.3.18"
 url = "2.5.2"
 uuid = { version = "1.10.0", features = ["v5"] }
 wiremock = "0.6.2"
-zitadel-rust-client = { git = "https://github.com/famedly/zitadel-rust-client", version = "0.3.0", default-features = false }
+zitadel-rust-client = { git = "https://github.com/famedly/zitadel-rust-client", package = "famedly-zitadel-rust-client", version = "0.12", default-features = false }
```
sirewix, line 36: We should use the version from crates.io

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+
+- *(zitadel)* Match Zitadel error codes across the full v2 error chain, restoring the "user already exists" update path and the invalid-phone fallback
+
+### Refactor
+
+- *(tests)* Migrate the e2e suite from the gRPC v1 client to the v2 HTTP API
+
+### Miscellaneous Tasks
+
+- *(deps)* Migrate to famedly-zitadel-rust-client v0.12 and drop the v1 feature, resolving the remaining security advisories
```
sirewix, line 17: It was only used in tests, so probably there was no real security issues

src/zitadel.rs
```diff
@@ -258,13 +261,19 @@ impl<'s> Zitadel<'s> {
 			}
 
 			Err(error) => {
+				// The v2 client wraps the actual Zitadel error (including the
+				// error code we match on below) as a source in the error chain,
+				// so we must inspect the full chain rather than just the
+				// top-level `Display`.
+				let error_chain = format!("{error:#}");
```
sirewix, line 268:
```suggestion
				let error_chain = format!("{error:?}");
```

src/zitadel.rs
```diff
@@ -371,14 +380,16 @@ impl<'s> Zitadel<'s> {
 
 		if let Err(error) = self.zitadel_client.update_human_user(zitadel_id, request.clone()).await
 		{
+			// The v2 client surfaces the Zitadel error code as a source in the
+			// error chain, so we match against the whole chain.
 			// If the new phone number is invalid
-			if error.to_string().contains("PHONE-so0wa") {
+			if format!("{error:#}").contains("PHONE-so0wa") {
```
sirewix, line 386:
```suggestion
			if format!("{error:?}").contains("PHONE-so0wa") {
```

src/zitadel.rs
```diff
@@ -371,14 +380,16 @@ impl<'s> Zitadel<'s> {
...
 				request.reset_phone();
 				self.zitadel_client.update_human_user(zitadel_id, request).await?;
 
 				if let Err(error) = self.zitadel_client.remove_phone(zitadel_id).await {
 					// If the user didn't start out with a phone
-					if !error.to_string().contains("COMMAND-ieJ2e") {
+					if !format!("{error:#}").contains("COMMAND-ieJ2e") {
```
sirewix, line 392:
```suggestion
					if !format!("{error:?}").contains("COMMAND-ieJ2e") {
```

tests/install-ids.rs
```diff
@@ -1,5 +1,7 @@
 //! E2E tests for the id installation script
 
+#![cfg(test)]
```
sirewix, line 3:
```suggestion
```

sirewix (commented):
Looks ok on the surface, but there are so many changes in test code. If we want to verify everything it will take a lot of time

Cargo.toml
nikzen: Done in 7fc2c4c: the dependency now points to crates.io (`zitadel-rust-client = { package = "famedly-zitadel-rust-client", version = "0.12" }`), with `Cargo.lock` resolving to the 0.12.2 release.

CHANGELOG.md
nikzen: Fair point \u2014 removed the security-advisory claim from the changelog in 7fc2c4c, since the legacy v1 client was only exercised by the test suite.

src/zitadel.rs
nikzen: Applied in 7fc2c4c.

src/zitadel.rs
nikzen: Applied in 7fc2c4c.

src/zitadel.rs
nikzen: Applied in 7fc2c4c.

tests/install-ids.rs
nikzen: Keeping `#![cfg(test)]`: without it the pinned-nightly clippy run (`cargo clippy --workspace --all-targets -- -D warnings`) fails this target with four `clippy::expect_used` errors, because the crate-level lint config in Cargo.toml warns on `expect_used` and CI runs with `-D warnings`. The attribute keeps the whole file out of the non-test compilation of the test target (the same pattern is already used in tests/e2e.rs and tests/common/mod.rs).

tlater-famedly:
We'll merge this as-is for now to resolve FP-295 - having spoken to Meike we'll probably end up reworking this all a *lot* with the new backend anyway, so I'm not too concerned as long as things continue to function in the short term.

We're already seeing breakage in production, so it's dubious that we should block this on testing.

tlater-famedly (approved):
See above comment

